PR #8, opened 2017-02-17
Basic pulsar class and working tests

jellis18:
So here is a basic `Pulsar` class that we can start to develop more along with `Signal` classes. I also have some very basic tests that work with the CI system.

Thinking about the design a bit more I think that the `Pulsar` class should stand on its own just as the `Signal` classes should stand on their own. Of course they will have to interact but they should be fairly independent. I haven't coded anything up yet but I think we could do something similar to how DMX is handled in PINT because DMX needs the TOAs to create bins but the DMX model class is still mostly separate from the TOA class.

At least we have some code now that works. Its a start...

@stevertaylor and @paulthebaker , take a look at this and then I'll pull the changes if we like this.


paulthebaker:
Can you set the future imports to:
```
from __future__ import (absolute_import, division, print_function, unicode_literals)
```

You currently just have:
```
from __future__ import division
```

While the more recent versions of Python 2.7 have the `print_function` not all versions do.  I don't think any have `absolute_import` or `unicode_literals` included.  While we probably won't have absolute/relative import clashes in our submodules or use unicode_literals, that is the recommended future import for full compatibility between versions.

[future module docs](https://docs.python.org/2/library/__future__.html)

jellis18:
@paulthebaker, that is a good point and I thought by having the future imports in the `__init__.py` it would take care of it but it looks like it doesn't. It seems like there should be a way to make it work that way instead of having to put that line in all of the modules. Do you know of a way to do that?

paulthebaker:
I'm pretty sure you just have to include it in every individual `.py` file...

jellis18:
OK, I've added the relevant future imports although I'm still not sure if I'm handling strings and unicode correctly but nothing fails so we are probably ok.

paulthebaker:
I think the problem is that all quoted literals default to unicode in Python 3, so when interacting with other modules like libstempo that expect string literals stuff breaks:
```

**Comment by jellis18** [comment]
@paulthebaker, that change casting of `epoch` to string should be in the most recent commit.

**Comment by paulthebaker** [comment]
They could also be declared as byte/string literals:
```
epoch = b'1950'
```

This might be what we want to avoid casting to string all the time.

paulthebaker:
I think this is ready to merge into master.  Any last changes or should I merge it?

jellis18:
Yep, its done!

PR #10, opened 2017-02-21
add .swp to gitignore

paulthebaker:
Vim creates tmp files named `.*.swp`.  Git shouldn't track them.

Also I'm testing pull requesting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# text editor tmp files
+.*.swp
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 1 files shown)


jellis18:
Hi @paulthebaker, yep this is a good idea and the PR worked, although I think the best way is to make the branch on your fork and then submit a PR from there.

paulthebaker:
> although I think the best way is to make the branch on your fork and then submit a PR from there

@jellis18, yeah that's what I'm doing with the prior PR... this was just a quick test...

PR #11, opened 2017-02-21
add basic Prior class

paulthebaker:
This adds a basic prior class that inherits `pdf` and `logpdf` methods from `scipy.stats` instances of the `rv_continuous` class.  For example:

```
from scipy.stats import norm
from enterprise.signals.prior import Prior
myPrior = Prior(norm(loc=0.0, scale=1.0))
print(myPrior.pdf(0.5))
```

There are a few convenience functions for uniform priors, but nothing fancy yet.


jellis18:
Can you make some unit tests for this? Once that is done we can merge this in I think.

paulthebaker:
I added `# noqa: E501` statements to the long URLs to manually ignore the `flake8` formatting errors.  I'm not sure how that will render in the auto-generated docs...

speaking of which, how do we view the auto-generated docs?

paulthebaker:
In debugging the unit tests I realized that the `a` and `b` parameters that define the domain of `rv_continuous` classes are defined against the base location and shape parameters.  To define a truncated Gaussian properly you would need to rescale the desired lower and upper bounds accordingly:
```
# A Gaussian prior centered on 0.9 and bounded to the domain [0, 1]
mean, std, lower, upper = 0.9, 0.1, 0.0, 1.0
a, b = (lower-mean)/std, (upper-mean)/std
myPrior = Prior(scipy.stats.truncnorm(loc=mean, scale=std, a=a, b=b))
```

I made a convenience function for uniform distributions to more naturally specify `lower` and `upper`.  I'm going to see if I can make a generic wrapper function that would do the same for any bounded `rv_continuous`.  If that is too much trouble I'll make a convenience function for a bounded Gaussian, as I expect it to be used.

jellis18:
> I made a convenience function for uniform distributions to more naturally specify lower and upper. I'm going to see if I can make a generic wrapper function that would do the same for any bounded rv_continuous. If that is too much trouble I'll make a convenience function for a bounded Gaussian, as I expect it to be used.

@paulthebaker, sounds good. I think once you get those changes and the tests pass (right now `make lint` is failing) we can merge it.

paulthebaker:
The docs don't format the way I want them to, but this is a working basic Prior class.  I vote to merge now and we can clean up the docs later.

jellis18:
OK, let me take a look and we can merge soon. I'll be needing this for my `Parameter` and `Signal` classes anyway!

paulthebaker:
I made the two long urls into references then put the `# noqa` statement in a sphinx comment at the end of the docstring, which seems to make it apply to the whole docstring.

jellis18:
Awesome! Are we ready to merge?

PR #14, opened 2017-02-24
Parameter and Signal classes

jellis18:
OK, I'm starting this PR so that we can discuss the `Signal` classes. I'm not quite sure how we should do this but first I'll just list the changes in this PR so far.

* Added basic `Parameter` class that basically just stores parameter names, values, and can evaluate the prior and can set new priors.
* Added Base Stochastic signal class. The methods will have to implemented by derived classes.
* Broke some PR rules and changed some docs settings (i.e. added math functionality, use ReadTheDocs theme, remove correct files when building docs)
* Setup [coveralls](https://coveralls.io) to `.travis.yml` to show test coverage in a nice way.
* Added tests for `Parameter` class.

So my problem is that I'm starting to think that having separate signal classes is a bit messy, or at least there should be a better way than what I have so far. In `StochasticSignal` there are several methods (i.e. `setup`, `add_param`, etc) that would be common to all signal classes and it seems wasteful to define them for every class. We could have a overall base class (like `TimingModel` in `PINT`) with these methods but that wouldn't be very nice either since the derived classes would need to have public methods that the parent class doesn't (like setting basis functions for stochastic signals and adding delays for deterministic signals). In `PINT` every timing model component basically just adds delays so this works.

Anyway, the goal is to have the `Signal` class very clean so that in the likelihood we could do something like
```
# s is initialized Signal class and psr is the pulsar class
s = Signals()
s.setup(psr)

# sets up parameters to the current parameter vector
s.init(params)

# get deterministic delays
s.get_delays()

# compute all gaussian process prior stuff
s.compute_phi_and_logdetphi()

# get white noise vector (tricky for ECORR)
s.get_white_noise()

# then compute likelihood from these signal components...
```

Lastly, even though we are focusing on single pulsar stuff at the moment we should always keep the full PTA in mind. Lets discuss this and see if we can get something good going.


vhaasteren:
What is wrong with derived classes that have public methods that the parent does not have? Seems very natural to me in this case. Especially if the stochastic class is general, with more specialized derived classes.

jellis18:
@vhaasteren, good to hear from you. I don't think there is anything wrong with it. I had a misunderstanding of the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle). Although now I think we may be leaning towards one base class which others are derived from. These derived classes will just not return certain elements. For example white noise or stochastic (at least the marginalized versions) will not return a delay (i.e. the delay subtracted off of residuals for deterministic signals). I think this may be the cleanest way. I imagine something like how `PINT` collects derivatives and various delays, not all derived classes return all different kinds of delays.

Anyway, @vallis had some ideas on Friday that he is going to post here soon. He suggested that we try to think of how we would like the user interface to be and then work backwards from there. @vhaasteren , do you have any suggestions?

vhaasteren:
I see. Yeah, with the LSP, the functions that use the abstract objects (unaware of the specific type) would only call the functions declared in the base class. But that does not mean the derived classes cannot have extra functions defined. So there is no problem. In fact, you should define extra functions in the stochastic class that all derived stochastic classes can then use in the same way. Most class hierarchies are built that way.

I do not have any specific suggestions, as I am kind of out of the loop. All I want to add at this point is that stochastic classes can and probably should have a delay. Remember, in the outlier code all the stochastic signals have the delays as model parameters (encoded as Fourier terms), so there the delay is accessible. In a hierarchical model all signals would work that way. Even though the white noise is then not treated that way, there is nothing to prevent one from doing it.

stevertaylor:
Building on what @vhaasteren said, maybe we could have a base signal class with delays, signal-basis functions, signal coefficients, and variances of these coefficients. For example, it is easy to write a CW signal or BWM signal with basis-functions and coefficients, even though we never do it and it is a bit contrived. If the coefficient variances of a signal are undefined then it can be treated as the usual deterministic signal, with delays but no phi matrix etc. If the variances are defined then a further option could be whether the likelihood is marginalized over the signal coefficients, which would dictate whether delays are computed for this signal.

jellis18:
@stevertaylor, yes I think that kind of the thing is the way to go. We can have a base class and then maybe a method or attribute that returns the signal type. I'll get on that later today and try to get some code up here. One thing that I will add and we will need is something like a "matrix multiplication function" that will just be `np.dot` for normal white noise but will use the Sherman-Morrison block diagonal stuff when ECORR is included in the white noise matrix. This will behave the same in that it will not be included for signals that don't need it.

paulthebaker:
I think all of this sounds good.  I think it's important that all of the class instances "look the same" to the basic functions that operate on them.

Steve's idea about writing deterministic signal in terms of basis functions and coefficients is nice.  Would the stochastic signals carry around zero-delays that could be added as usual?  This wastes operations, but might simplify the code to avoid type some checking.

I agree with Justin that we don't want to have to choose which matrix multiply to use based on which signal we're talking about.  The "matrix multiply" function should have access to some metadata that comes with the input to decide which method to use.

paulthebaker:
We should probably look into the rebase conflict now and get this realigned with the master branch now, before adding too many new changes.

jellis18:
@paulthebaker, what rebase conflict?

paulthebaker:
right above the comment box on this page:

> **This branch cannot be rebased due to conflicts**
>Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

paulthebaker:
It appears to be in `requirements_dev.txt`

run `git rebase nanograv/master` or whatever you call the main tracked branch

stevertaylor:
@paulthebaker The stochastic signals might still have a delay property, but it would be set to None unless the likelihood was set to be fully hierarchical over this signal. If the likelihood is marginalized over the signal then we don't need to create the delay.

jellis18:
OK, So after discussion with @vallis and @stevertaylor we are thinking of probably having a base `Signal` class and then derived classes for red noise, deterministic, white noise, etc. @vallis and I are playing around with the implementation of those but we were wondering what the best way was to instantiate such a class. Say we have a class `FourierBasisNoise` that inherits from `Signal`, we want to be able to have different kinds of PSD here and we would like to  just have a `spectrum` option in the initializer. This option may be the simplest but then you get into the problem of having a lot of if statements in the class and I'd really like to avoid that. So I came up with a nice pattern to at least define and register different PSD's, we could potentially use this for the whole signal class but I haven't though of that yet, at the very least we could just use it for red noise like signals. I have some code below that you should be able to run
```
from __future__ import (absolute_import, division,
                        print_function, unicode_literals)

import numpy as np
from enterprise.signals.parameter import Parameter
import matplotlib.pyplot as plt

class PSD(object):
    
    _registry = {}
    
    def __new__(cls, name, *args, **kwargs):
        psd = super(PSD, cls).__new__(cls, name, *args, **kwargs)
        cls._register(psd, name)
        return psd
    
    def __init__(self, name):
        self.params = []
    
    @classmethod
    def _register(cls, psd, name):
        cls._registry[name.lower()] = psd
        psd._name = name.lower()
        
    @classmethod
    def names(cls):
        return cls._registry.keys()
        
    @property
    def name(self): return self._name

    @classmethod
    def get(cls, name):
       
        # Be case-insensitive
        name = name.lower()
        # First see if name matches
        if name in cls._registry.keys(): 
            return cls._registry[name]
        # Nothing matched, raise an error
        raise KeyError("PSD name '%s' is not defined" % name)
        
    def add_param(self, param):
        setattr(self, param.name, param)
        self.params += [param]
        
    def get_psd(self, f):
        """Must be defined in derived class"""
        raise NotImplementedError
        
class PowerLawSpectrum(PSD):
        
    def __init__(self, name):
        super(PowerLawSpectrum, self).__init__(name)
    
        self.add_param(Parameter('A', -15.0, uncertainty=0.1,
                                 description='Red Noise Amplitude [log-scale]'))
        self.add_param(Parameter('gamma', 4.33, uncertainty=0.1,
                                 description='Red Noise Spectral index'))
    
    def get_psd(self, f):
        df = f[0]
        f1yr = 1 / 3.16e7
        Amp = 10**self.A.value
        gamma = self.gamma.value
        psd = Amp ** 2 / 12 / np.pi ** 2 * f1yr ** (gamma - 3) * f ** (-gamma) * df
        return psd
 
# call this to register (could go in an __init__.py)      
PowerLawSpectrum('powerlaw')

# test
psd = PSD.get('powerlaw')

T = 11 * 3.16e7
f = np.linspace(1/T, 50/T, 50)

plt.loglog(f, psd.get_psd(f))

# change the value
psd.A.value = -15.5
plt.loglog(f, psd.get_psd(f))
```

So in this case a new user would not have to modify any existing code but only create a new class derived from PSD and define the parameters and psd and then add it to the registry list (i.e., the thing in the `__init__.py`). Let me know if you all think this is useful. Of course it would need some tweaks but I like it.

P.S. After writing this I'm thinking more and more that we could probably use this kind of pattern for the whole signal class...

stevertaylor:
So @jellis18 are you thinking that `FourierBasisNoise` lives above `PSD` in the class hierarchy? On the same level as `PSD` you could have `FourierSeries` which would be the actual time series of the process in the hierarchical likelihood case. Like what @paulthebaker suggested, you could give `FourierBasisNoise` an option for `marginalized=True/False`, where if `True` then the output of `FourierSeries` is just an array of zeroes of length Ntoas.

paulthebaker:
@jellis18, this is great!  We should probably pass the amplitude and spectral index and their uncertainties at declaration:
```python
class PowerLawSpectrum(PSD):
    def __init__(self, name, logA=-15., dlogA=0.1, gamma=4.33, dgamma=0.1):
...
```

paulthebaker:
I also tried using a Gaussian prior on `A` and `gamma`.  In the `PowerLawSpectrum.__init__()` declaration you can pass a prior to `add_param`
```python
from scipy.stats import norm

...

self.add_param(Parameter('A', logA, uncertainty=dlogA,
                         prior=prior.Prior(norm(loc=logA, scale=dlogA)),
                         description='Red Noise Amplitude [log-scale]')
               )
```

It all works nicely. Try
```python
print("A={0:.1e},  logPrior(A)={1:.3f}".format(10**psd.A.value, psd.A.prior_pdf()))
```
or
```python
logArange = np.linspace(-18., -12., 100)
plt.semilogx(10**logArange, psd.A.prior_pdf(logArange, logpdf=False))
```
image

jellis18:
@stevertaylor, the idea was to have `PSD` as a registry of PSDs that carry along useful methods so that they could be initialized within the `Signal` class. That way you could define something like
```python
class FourierSeries(Signal):
    def __init__(self, psd='powerlaw', nf=30, marginalize=True):
        ...
        self.psd = PSD('powerlaw')
        ...
```
So in the end PSD is not directly in the hierarchy but lives as a separate registry class that is used here because there are so many different types of PSDs that one can use and we would like to avoid separate signals for every one and to also avoid a ton of `if` statements.

jellis18:
@paulthebaker 

> `PowerLawSpectrum('powerlaw', logA=-15., dlogA=0.3, gamma=4.333, dgamma=2/3)`

That doesn't work because in the PSD class it is setup to only take name and register it based on the name. Although now that I look at this code, it isn't going to work for multiple pulsars because it (at least as it is now) only creates one instance of the PSD class so if different pulsar called `PSD.get` then it would just return the same object. Maybe I got too excited with this registry pattern and I need to think a bit more.

vallis:
An example of the metaclass pattern that I discussed with Justin and Stephen. We define basic reusable mathematical objects as customized class, then instantiate them when they are applied to a particular pulsar.

```python
# I didn't have time to study your parameter implementation, so this is my own naive version

class Parameter(object):
    def __init__(self,name):
        self.name = name

# the following two are class *factories*: they return a customized class
# then the actual parameters (with names) are obtained by instantiating the class

def Uniform(pmin,pmax):
    # this naive class definition can be improved with type, etc.
    # but will be good enough for a demo
    class Uniform(Parameter):
        def pdf(self,p):
            return 1.0 / (pmax - pmin) if (pmin < p < pmax) else 0.0
                
        def __repr__(self):
            return '"{}":Uniform({},{})'.format(self.name,pmin,pmax)
        
    return Uniform

def Normal(mu=0,sigma=1):
    class Normal(Parameter):
        norm = 1.0 / math.sqrt(2 * math.pi * sigma**2)
        
        def pdf(self,x):
            return self.norm * math.exp(-0.5*(x - mu)**2/sigma**2)
        
    return Normal

# a uniformly distributed parameter
u = Uniform(0,10)

u
[out:] __main__.Uniform.<locals>.Uniform

# instantiations of this parameter
u1 = u('uniform1')
u2 = u('uniform2')

u1
[out:] "uniform1":Uniform(0,10)

u1.pdf(5), u1.pdf(-1)
[out:] (0.1, 0.0)

n = Normal(2,5)

n1 = n('normal1')
n2 = n('normal2')

n1.pdf(2)
[out:] 0.07978845608028654

# base signal class. Not much in it yet
class Signal(object):
    @property
    def params(self):
        return self._params.values()

# again, a factory that returns a customized class, reusable to generate
# signal instances for many pulsars

def MeasurementNoise(efac = Uniform(0.5,1.5)):
    class MeasurementNoise(Signal):
        def __init__(self,psr):
            self._psr = psr
            
            self._params = {'efac': efac(psr.name + '_efac')}
            self._ndiag = (psr.toaerrs*1e-6)**2

        # here we assume that we pass all parameters to all the "tensor"
        # methods of the instantiated signals. An alternative is to set
        # a persistent value for each parameter. It's less functional,
        # but (often) practicality beats purity
        def ndiag(self,params):
            return params[self._params['efac'].name]**2 * self._ndiag
        
    return MeasurementNoise

w = MeasurementNoise()

psr = libstempo.tempopulsar('../2222.par','../2222.tim')
w1 = w(psr)

w1.params
[out:] dict_values(["J2222-0137_efac":Uniform(0.5,1.5)])

w1.ndiag({'J2222-0137_efac': 1.0})[0]
[out:] 3.7423676303999994e-10

w1.ndiag({'J2222-0137_efac': 1.5})[0]
[out:] 8.4203271683999982e-10

x = MeasurementNoise(efac=Normal(1,0.5))
x1 = x(psr)

x1.params
[out:] dict_values([<__main__.Normal.<locals>.Normal object at 0x1164d1fd0>])
```

paulthebaker:
@jellis18, within the `Parameter` class, what will `uncertainty` be used for?  So far, nothing.

Shouldn't that info be in the `Prior` at initialization?  It's not entirely clear what "uncertainty" means for different priors anyway.  If you want to access something like uncertainty, couldn't you query the prior or posterior for a credible interval as needed?

jellis18:
@paulthebaker 

As we are going now there won't be a `Parameter` class as there is now. It will just be defined how we have it in the development notebooks. The idea before was just to have a basic idea of how much you expect the parameter to vary so that you can use it to set initial jump proposals.

jellis18:
I'm going to close this as it has somewhat been overtaken by the notebooks and the new PR #54.

PR #17, opened 2017-03-01
Fixed some errors in point 3 of Get Started in CONTRIBUTING.rst

stevertaylor:
When following the guidelines in CONTRIBUTING.rst, I had to add recursive options before the requirements_dev.txt and requirements.txt files. Also, `make test` would not run without first installing `libstempo` in my virtual environment.

PR #18, opened 2017-03-01
Get docs to auto-generate with ReadTheDocs

jellis18:
Steve's PR #17, reminded me that the docs were not auto-generating the docstrings for the python code. This is actually a common problem and I think I've come up with a way to fix it. With my tests it appears to work now. 

I did have to add a `try-except` statement to `pulsar.py` because `sphinx` needs to be able to import your module. There may be a better way around this but it works for now without changing any functionality.

To test this I made my own RTD docs page for my for and you can see it [here](http://enterprise-dev.readthedocs.io/en/latest/) and you can see that the the modules are all listed as they should be.


jellis18:
I'll probably just merge this in a little while since it really doesn't effect any code but simply makes things more automated.

PR #20, opened 2017-03-02
constants module

paulthebaker:
A module containing useful constants for consistency across modules.

### DIFF
--- a/enterprise/constants.py
+++ b/enterprise/constants.py
@@ -25,6 +25,7 @@ h = sc.Planck
 
 # astronomical times in sec (and frequencies in Hz)
 yr = sc.Julian_year
+day = sc.day
 fyr = 1. / yr
 
 # astronomical distances in meters
@@ -34,6 +35,12 @@ pc = sc.parsec
 kpc = pc * 1.e3
 Mpc = pc * 1.e6
 Gpc = pc * 1.e9
+
+# solar mass in kg and m,s natural units
+GMsun = 1.327124400e11  # measured more precisely than Msun alone!
+Msun = GMsun / G
+Rsun = GMsun / (c**2)
+Tsun = GMsun / (c**3)
 
 # other useful conversions in mks
 erg = sc.erg
--- a/enterprise/constants.py
+++ b/enterprise/constants.py
@@ -44,3 +44,6 @@ Tsun = GMsun / (c**3)
 
 # other useful conversions in mks
 erg = sc.erg
+
+# other things
+DM_K = 2.41e-16  # for DM variation design matrix
--- a/enterprise/constants.py
+++ b/enterprise/constants.py
@@ -1,0 +1,39 @@
+#constants.py
+"""Declares physical constants for use in enterprise.
+Depends on numpy for base mathematical constants, and
+scipy.constants for physical constants.
+"""
+
+from __future__ import (absolute_import, division,
+                        print_function, unicode_literals)
+
+import numpy as np
+import scipy.constants as sc
+
+
+# mathematical constants from numpy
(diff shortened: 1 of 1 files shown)


paulthebaker:
The main question is: **Do we want this?**

I think it's useful to have things like `fyr` declared once, so they are the same everywhere.  All of the physical constants are in scipy.constants anyways.  This is just for convenience.

jellis18:
I think we would want this, especially when we get to continuous GWs and we want to get rid of all the units. We should definitely include things like the number of seconds per day and maybe include conversions like sec_to_day?

paulthebaker:
@jellis18 By the current convention `yr` or `day` or `erg` is expressed in mks units.  In my mind it makes the standard conversion more readable i.e.
```python
 5 * constants.yr  # 5 yr in sec
10**49 * const.erg  # SN energy in mks
```
I like it this way, but I wouldn't be strongly opposed to renaming things to make the conversions more explicit.

I included a solar mass expressed in each mks as `Rsun`, `Msun`, and `Tsun`.  I anticipate confusion with `Rsun`... This has always been my naming convention, so it makes sense to me.  Toes anyone have a different naming convention they like?

stevertaylor:
@paulthebaker Could you maybe add one more constant? It's `K = 2.41e-16` used in constructing design matrices for the Gaussian-process DM-variation model.

paulthebaker:
We should have a standard convention for using the constants module.  I suggest
```python
import enterprise.constants as const
```

and **NOT**
```python
from enterprise.constants import *
```

stevertaylor:
Agreed. We need to know which module all functions and attributes are coming from.

paulthebaker:
@stevertaylor What should I call it?  I don't think `K` is descriptive enough.  Is `DM_K` okay?

scottransom:
Any reason why you wouldn't want to use the super-nice units and constants from astropy?  Besides the small amount of efficiency hits you get (which I don't think should be an issue here given that most of the time will be spent in matrix calculations, I believe...)

stevertaylor:
@paulthebaker That sounds fine to me. Thanks.

jellis18:
@scottransom, that might be a good idea but other than me getting really confused with PINT units I don't think we have much experience with it...

paulthebaker:
@scottransom, I considered it.  I think we would then need to carry around units on all of our parameters.  I'm personally used to picking a system of units, converting everything to that and sticking to it.  For GW stuff that usually means converting everything natural units of seconds.

Neither `PAL2` nor `NX01` used units, but now would be the time to start.  If everyone wants units we can do it.  My vote would still be no units.

paulthebaker:
@scottransom @jellis18 I also get very confused when dealing with libraries with units...

scottransom:
Well you are doing your own units thing now!  And the nice thing about astropy units is that they allow you to convert as need to whatever you want.  And it is much easier to prevent consistency issues.

Just curious, @jellis18 what confused you about the units in PINT?  I think they make the code much easier to read / and write!

scottransom:
Note also that I really think you should be considering moving the analysis code to using PINT in the future.  TEMPO2 is in a scary state, IMO.  And if PINT becomes the future of these types of analysis, then 1) units come for "free" with it and 2) libstempo becomes less and less important (since you can work with the PINT stuff directly).  I think these are important things to consider.

paulthebaker:
@scottransom Yes!  We definitely want `PINT`.  We started with `libstempo` because that's what the old codes used.  We had some plans to be able to work with either.  But I agree that `PINT` is the future.

For now I'm going to merge this as is to move to the basic working code goal.  When we look at using `PINT` we will return to this discussion.

PR #21, opened 2017-03-02
Starting a utilities module

stevertaylor:


jellis18:
Here is a somewhat shorter way to create Fourier design matrices.
```
def create_fourier_designmatrix(t, nmode=50):
    
    # initialize
    F = np.zeros((len(t), 2*nmode))
    
    # set total time
    T = (t.max()-t.min())
    
    # fourier spaced frequencies
    f = np.linspace(1/T, nmode/T, nmode)
    
    F[:,::2] = np.sin(2*np.pi*t[:,None]*f[None,:])
    F[:,1::2] = np.cos(2*np.pi*t[:,None]*f[None,:])
    
    return f, F
```

PR #24, opened 2017-03-04
Further utilities functions

stevertaylor:


paulthebaker:
@stevertaylor can you use the `enterprise.constants` module instead of defining your own constants.

paulthebaker:
I don't think having a single, sprawling `utils` module is the best way to go.  I think we should break these out into smaller subsets of functions that go together and put them in the submodule that will use them, as like `enterprise.signals.utils` or `enterprise.likelihood.utils`.  Or maybe we should name them more meaningfully.  I want to consider how other users will most easily find the functions they need.

jellis18:
@paulthebaker, I agree with that too. We will have to think about how we want to break them up though...

stevertaylor:
@paulthebaker Yes, thinking about it, we definitely need to break these up. The problem is that the list so far is still only signals, and there are more to go...

PR #29, opened 2017-03-08
Adding epoch and system sorting to pulsar.py

stevertaylor:


stevertaylor:
The CI may fail because the sorting needs some functions in likelihood utils, which has not been merged from my other branch yet. I'm just opening this pull request to show you all where the sorting is going to happen.

jellis18:
@stevertaylor, this is good. One thing that we will have to do is to potentially make a larger chain of flag choices. By that I mean that some datasets do to have a `group` of an `f` flag, the NANOGrav 5-year data for instance. I'll try to think of a cleaner way of doing that other than a large chain of if statements as I have now.

Also, related to this is that we should probably specifically make a property that returns the backend flags that were used in this sorting so that it can be used later in making parameters and masks for EFAC, EQUAD, and ECORR.

jellis18:
I think we can probably get rid of this

PR #54, opened 2017-05-02
Base code for pulsar signals

jellis18:
I'm starting this PR so that we can all see the direction that we are going. This PR is not ready to merge as I don't have any tests and the doc strings are pretty sparse. So of course we still need to nail down the multiple pulsar (spatially correlated analysis) but I think this should be merged first. Take a look and let me know what you think. I will get working on some tests...


jellis18:
OK, I have changed things a bit to be more flexible in terms of how we select indices for parameters. Previously if we wanted to do EFAC, EQUAD per backend then I had the `by_backend` flag for the class factories. However, that hardcodes that behavior and we would have to define another factory to do something different.

So to try to generalize this I've started a `Selection` factory, similar to how we use the `Function` factory to give arbitrary PSDs. The goal here is to make it so that when a 'Selection' class is passed to the signal, it will initialize with the pulsar and parameter and will automatically generate the masks required to compute the signal (i.e. which indices correspond to which backend or which time-segment). This is a bit tricky if we want this to not have information about the pulsar when we define the `Selection` class. At the moment I've done this by requiring that the arguments to the mask function (i.e the custom function that provides the mask for the parameters) be part of the pulsar object. However for things like ECORR we want to use epoch-averaged versions of things like the flags or toas, which are currently not part of the pulsar object. We could add them but I worry that this may lead to problems down the road when we want some complicated filtering.

Here is some code for how I imagine the API looking (the ECORR does not have this yet):
```python
# pulsar information
parfiles = 'data/B1855+09_NANOGrav_11yv0.gls.par'
timfiles = 'data/B1855+09_NANOGrav_11yv0.tim'

psr = Pulsar(parfiles, timfiles)

# white noise parameters
efac = parameter.Uniform(0.5, 5)
equad = parameter.Uniform(-10, -5)
ecorr = parameter.Uniform(-10, -5)

# selection object to use parameters "per backend"
selection = Selection(utils.get_flag_masks)

# another selection to just naively have one two EFACS for the first and second half of data
def cut_half(name, toas):
    mid = (toas.max() + toas.min()) / 2
    ind = [toas <= mid, toas > mid]
    labels = ['t1', 't2']
    return [('_'.join([name, label]), ii) for ii, label in zip(ind, labels)]

selection2 = Selection(cut_half)

# EFAC and EQUAD per backend
ef = ws.MeasurementNoise(efac=efac, selection=selection)
ef2 = ws.MeasurementNoise(efac=efac, selection=selection2)
eq = ws.EquadNoise(log10_equad=equad, selection=selection)

# still need to figure out selection for ECORR
ec = gs.EcorrBasisModel(log10_ecorr=ecorr, by_backend=True)

# red noise
pl = Function(utils.powerlaw, log10_A=parameter.Uniform(-18,-12), gamma=parameter.Uniform(1,7))
rn = gs.FourierBasisGP(spectrum=pl, components=30)

# timing model
tm = gs.TimingModel()

# combined signal
s = ef + ef2 + ec + rn + eq + tm

# initialize
pm = s(psr)
```

I think this is really cool but it still needs some work. @vallis, you are good at this kind of thing. Want to take a look.

BTW, this functionality does not need to be limited to white noise, we could also use this to filter the rows of the F-matrix for "system" type noise.

jellis18:
Ok @vallis and I have simplified the selection function even more so that any user defined selections are very minimal. EFAC, EQUAD, and ECORR can now take any kind of selection. I still have to work on the `FourierBasisGP` class and get selection working with that (it is a bit different because we use the `Function` class for the parameters and spectrum). Getting closer!

jellis18:
So we are going to have problems with ECORR as it is coded now because we are using dictionaries which have no fixed order, so concatenating arrays based on the keys (as in the ECORR model) will not work. I'm mainly just putting this here so we remember that this needs to get changed...

vallis:
Re dicts, perhaps OrderedDict is our friend here.

jellis18:
I think it is all taken care of now with normal dicts. I'll know for sure when I start writing tests!

jellis18:
Ok with the last commit, I think things are good to go. I have removed @vallis's changes as that should go in a different PR. I have written **many** (we are up to 87% coverage now!) tests for the signals and I'm 99.9% certain that everything is working fine. If you can think of any other tests to add now let me know.

jellis18:
Ok so we aren't quite done. It is probably best to just make this PR have the full pulsar signal, including multiple pulsars. So more tests!

jellis18:
With the last commit we at least have some coverage of the `PTA` class. There are some other scenarios that we can test in this PR but I don't want to go into huge detail since some of the underlying implementation (i.e sparse matrices, generic solve methods etc) will change in other PRs.

jellis18:
Ok, now that the tests finally pass I think we can merge this so that we have a stable base to work from in the future.

PR #89, opened 2017-05-16
Add PTA likelihood, fix a few tests for Python 3

vallis:
At first sight, seems to work... however, for certain subsets of 4+ pulsars from the 11yr release dataset, the Cholesky inversion of Phi fails (matrix not positive definite). Maybe the matrix is too sparse, or perhaps the numerical values I'm trying are the problem. Or a bug.

According to my timings, the following may be the fastest way to sum the sparse phiinv and the TNT sequence. It's also the prettiest. However I'd like to time more pulsars, and I can't because of Cholesky.

```python
class MarginalizedLogLikelihood3(signal_base.MarginalizedLogLikelihood):
    def _make_sigma(self,phiinv,TNTs):
        return phiinv + sps.block_diag(TNT,'cdc')
```


vallis:
I pass, yay, I pass!

jellis18:
@valis, you have passed the easy part. Now to write some tests for the Likelihood!

vallis:
Is there an automatic coverage requirement? (Not as a PR merge hook...)

jellis18:
The coverage can't be below 80% but this module **really** needs tests because it is the central part.

jellis18:
@vallis, you can run `make coverage` to test the coverage on your machine.

vallis:
Ok, shall test. It's perversely fun.

M

vallis:
I've managed to come up with a little diabolical algorithm that inverts phi matrices by frequencies even in the presence of multiple correlated processes that share the same Fourier basis vectors. It's based on the idea of "clique" (a set of phi-matrix indices that share non-diagonal terms).

- My limited testing shows that I reproduce the output of the sparse algorithm.
- The determinant is slightly off. Is it a loss of precision, or a subtle bug?
- You can now get the three algorithms by doing get_phiinv(params,method), with method in ['cliques', 'partition', 'sparse']. Note that 'partition' is only correct for a single correlated process.
- For 10 pulsars, 20 frequencies, and 2 processes, 'cliques' is 2x faster than 'sparse', and 20% slower than 'partition'.

jellis18:
Ok, you may have to go over this with me in person. Although, it is possible to make the 'partition' method work for multiple spatially correlated signals given that the frequencies are the same? I imagine that would almost always be the case but with 'cliques' it doesn't have to be.

jellis18:
Since this is **very** important base code functionality we want to have very comprehensive tests. I would suggest (all with efac, rn, tm, and common signals):
1. Compare all phiinv methods to analytic solution
    * 1 common signal, same numbers of frequencies for all 
    * 1 common signal, different numbers of frequencies for intrinsic RN and common signal
    * 1 common signal, different frequencies (i.e. different Tspan) for intrinsic RN and common signal
2. Compare `cliques` method to analytic solution
    * 2 common signals, same numbers of frequencies for all 
    * 2 common signals, different numbers of frequencies for intrinsic RN and common signals (same for common signals)
    * 2 common signals, different numbers of frequencies for intrinsic RN and common signals (also different numbers of frequencies for common signals)
    * 2 common signals, different frequencies (i.e. different Tspan) for intrinsic RN and common signal (same for common signals)
    * 2 common signals, different frequencies (i.e. different Tspan) for intrinsic RN and common signal (also different Tspan for different signals)

jellis18:
@vallis, I've just submitted a PR to your branch implementing caching. I can't seem to push directly here.

jellis18:
Lets do it!

Issue #90, opened 2017-05-18
How do we subclass signals?


jellis18:
Currently all signals are defined by class factories of the form 
```python
def SignalName(*args, **kwargs):
     class SignalName(Signal):
         def __init__(self, psr):

    # define things here

    return SignalName
```
The question is how can we subclass the specific signal class? For example we have `FourierBasisGP` signals for red noise but a `DMFourierBasisGP` factory would be identical except the `F` matrix and parameter names would be different. We **definitely** need some way of doing this, otherwise we are going to be re-writing a lot of code.

Any thoughts?

vallis:
This works:

```python
def DerivedSignalName(*args, **kwargs):
     class DerivedSignalName(SignalName(*args,**kwargs)):
          # override methods here

     return DerivedSignalName
```

Is it confusing?

jellis18:
@vallis, this is good but we may need to re-work some of the signals to be modular enough so that we minimize re-writing of code when sub-classing.

jellis18:
I think this is good now. I'm closing this issue.

PR #91, opened 2017-05-19
New ECORR signal models

jellis18:
I'm starting this PR, again, so we can take a look at how things are progressing. It still needs tests and some cleanup related to Issue #90. Some of the new code is pretty cool, IMHO. To deal with ECORR in the `N` matrix we need to make it play nice with the standard `N` "vector" (i.e. the diagonal part of a diagonal matrix) that we use for EFAC, and EQUAD. In order to do this I've created three special data types: `ndarray_alt`, `csc_matrix_alt`, and `ShermanMorrison`. They all have their own `__add__` methods so that they know how to add to one another (note that `csc_matrix_alt` cannot be added to `ShermanMorrison` at the moment but I don't know if you would ever want to do that anyway). They also have their own `solve` methods that have the same API for all so that the `Likelihood` doesn't need to know which kind of `N` you are using.

Here is how it works: the EFAC and EQUAD signals return `ndarray_alt` arrays (basically `numpy` arrays with special `__add__` and `solve` methods), ECORR (sparse) signals return `csc_matrix_alt` arrays (which, again, are a special type of `spicy.sparse.csc` matrix with `__add__` and `solve` methods), and ECORR (sherman-morrison) signals return a `ShermanMorrison` array. This way when computing the total `N` in `SignalCollection` using the `sum` function we return the correct type. For example `ndarray_alt` + `csc_matrix_alt` returns a `csc_matrix_alt` type array where we can then call `solve` in the likelihood. The same is true for `ndarray_alt` + `ShermanMorrison` where it returns a `ShermanMorrison` type array. I have done tests and all of this works smoothly but there are still a few more checks to do.

Here is how to use the solve method (for any of the three types mentioned above):
To solve Y^T N^{-1} X, where X and Y can be either vectors or matrices of the correct shape you can do `N.solve(X, left_array=Y)`. 

Lastly, in terms of the ECORR signal factories themselves I have defined `EcorrKernelNoise` which implements the sparse matrix version and have sub-classed that to create `EcorrKernelNoiseSM` which implements the Sherman-Morrison version. I would still like to clean these up a bit and make them easier to subclass but they are working now.

Timings (model `m` is EFAC + ECORR(sherman morrison) + timing model, model `m2` is the same but with ECORR (sparse)) for J1909-3744:
```python
T = m.get_basis()
N1 = m.get_ndiag(params)
N2 = m2.get_ndiag(params)
%timeit -n5 m.get_ndiag(params)
%timeit -n5 m2.get_ndiag(params)
%timeit -n5 N1.solve(psr.residuals, left_array=T, logdet=True)
%timeit -n5 N2.solve(psr.residuals, left_array=T, logdet=True)
%timeit -n5 N1.solve(T, left_array=T)
%timeit -n5 N2.solve(T, left_array=T)
```
returns :
```python
5 loops, best of 3: 159 µs per loop
5 loops, best of 3: 99.7 ms per loop
5 loops, best of 3: 10.5 ms per loop
5 loops, best of 3: 29.9 ms per loop
5 loops, best of 3: 86 ms per loop
5 loops, best of 3: 96.6 ms per loop
```
So the solve method is fairly similar for the two methods but constructing `N` for the sparse case is very slow...

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -24,7 +24,7 @@ except ImportError:
 class Pulsar(object):
 
     def __init__(self, parfile, timfile, maxobs=30000, ephem=None,
-                 planets=True):
+                 planets=True, sort=True):
         # Check whether the two files exist
         if not os.path.isfile(parfile) or not os.path.isfile(timfile):
             msg = 'Cannot find parfile {0} or timfile {1}!'.format(
@@ -43,6 +43,9 @@ class Pulsar(object):
         # Change directory to the base directory of the tim-file to deal with
         # INCLUDE statements in the tim-file
         os.chdir(dirname)
+
+        # sorting
+        self._sort = sort
 
         # Load pulsar data from the libstempo library
         # TODO: make sure we specify libstempo>=2.3.1
@@ -176,6 +179,20 @@ class Pulsar(object):
             self._planetssb[:, 7, :] = self.t2pulsar.neptune_ssb
             self._planetssb[:, 8, :] = self.t2pulsar.pluto_ssb
 
+        # sorting
+        self.sort_data()
+
+    def sort_data(self):
+        """Sort data by time."""
+        if self._sort:
+            self._isort = np.argsort(self._toas, kind='mergesort')
+            self._iisort = np.zeros(len(self._isort), dtype=np.int)
+            for ii, p in enumerate(self._isort):
+                self._iisort[p] = ii
+        else:
+            self._isort = slice(None, None, None)
+            self._iisort = slice(None, None, None)
+
     def filter_data(self, start_time=None, end_time=None):
         """Filter data to create a time-slice of overall dataset."""
         if start_time is None and end_time is None:
@@ -199,6 +216,8 @@ class Pulsar(object):
         if self._planetssb is not None:
             self._planetssb = self.planetssb[mask, :, :]
 
+        self.sort_data()
+
     def to_pickle(self, savedir):
         """Save object to pickle file."""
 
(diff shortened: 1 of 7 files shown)


jellis18:
Ok, I've added new type for ECORR, `BlockMatrix`. It is similar to `ShermanMorrison` but takes a list of blocks (i.e., the small arrays that make up the block diagonal) and solves systems by just inverting the individual blocks. This way it is generic enough to handle blocks that cannot be written as the outer product of two vectors (a requirement for `ShermanMorrison`).  Oddly enough it is actually faster than the sparse matrix version overall. 

Overall it still works but I'm not quite happy with how I've organized this. Right now there are three separate class factories for the three different methods of doing ECORR. This may not be the right thing to do as the signals are identical it is just the method that changes. Doing things this way does avoid having a bunch of if statements in the class if we were to try to include the method as an argument but maybe that is preferable in this case.

Thoughts?

jellis18:
I think this is ready to merge.

jellis18:
Ok, now I think we are good. After talking with @vallis, we've decided it is cleaner to just have one class factory with a `method` argument. It does involve a few internal methods and some if-statements but it is much clearer overall and makes more sense since these are not different signals, merely different methods of constructing the same signal.

So the signal is initialized as follows
```python
# method can be sherman-morrison, block, or sparse
ec = EcorrKernelNoise(log10_ecorr=Uniform(-10, -5), selection=selection, method='sherman-morrison')

# initialize efac (ecorr cannot really act on its own so we always need efac even if fixed to 1)
ef = MeasurementNoise(efac=Uniform(0.5,1.5),  selection=selection)

s = ef + ec
model = s(psr)

# then the signal collection knows how to handle block-diagonal N-matrices for any of the three methods
N = model.get_ndiag(params)

# matrix operations are done with the following

# T^T N^{-1} T
TNT = N.solve(T, left_array=T)

# T^T N^{-1} \delta t
TNr = N.solve(psr.residuals, left_array=T)

# \delta t^T N^{-1} \delta_t
rNr = N.solve(psr.residuals, left_array=psr.residuals)

# the solve method also takes a boolean logdet option to return the log-determinant
```

jellis18:
Ok, for whatever reason there is some black magic happening when using `.. math::` in the travis tests. For now I have just removed an equation from the docstrings. We can deal with this more when we finalize the documentation.

Merging now.

Issue #93, opened 2017-05-25
How to do generic Fourier signals?


jellis18:
We already have the ability go pretty much use whatever spectrum we want through the `FourierBasisGP` class factory and we can do lots of different selections. However, right now the basis is fixed (i.e. for something like DM we would need a completely new signal factory or at least a lot of copied code).

I've been working on a  few things to make this more flexible and I'd like to know what you think. I've defined a new factory (the name could change):
```python
def Fourier(func, ftype, **func_kwargs):
    
    class Fourier(object):
        
        def __init__(self, psr):
            self._psr = psr
            
        def __call__(self, **kwargs):
            kwargs.update(func_kwargs)
            return selection_func(func)(self._psr, **kwargs)
        
        @property
        def ftype(self):
            return ftype
    
    return Fourier
```
This uses some of the fancier stuff that we are using from the selection, that is, if the underlying `func`s arguments are attributes of the pulsar class then it will use them. This way we can still use generic functions without having to specifically write them to read in the `Pulsar` class (just make the arguments attributes of the `Pulsar` class).

So this could interact with `FourierBasisGP` as follows
```python
pl = base.Function(utils.powerlaw, log10_A=parameter.Uniform(-18,-12), gamma=parameter.Uniform(1,7))
basis_dm = Fourier(utils.createfourierdesignmatrix_dm, 'dm')
basis_red = Fourier(utils.createfourierdesignmatrix_red, 'red')
rn = FourierBasisGP(pl, basis_red, selection=selection)
dm = FourierBasisGP(pl, basis_dm, selection=selection)
```
for DM  and red power-law noise. Basically it would now require a spectrum and a basis. The `type` part of it is needed so that we could have different kinds. The cool thing is that we can also set arguments on the outside so if we wanted  like ` Fourier(utils.createfourierdesignmatrix_red, 'red', fmin=1e-9, fmax=1e-7)`.

What do you all think?

jellis18:
So @vallis wrote on Slack the other day

> I think the best approach is to decide what API we want, then figure out how to implement it. Thinking very generally, maybe we want an abstraction like `BasisGP(priorFunction, basisFunction, selectionFunction)`

[...]

PR #95, opened 2017-06-08
BasisGP and BasisCommonGP

jellis18:
I have made some generic `BasisGP` classes that go along with the discussion in Issue #93. The `BasisGP` and `BasisCommonGP` functions looks like
```python
def BasisGP(priorFunction, basisFunction,
            selection=Selection(selections.no_selection),
            name='')

def BasisCommonGP(priorFunction, basisFunction, orfFunction, name='common')
```
Where the arguments use the new and improved `Function` class that can take parameter arguments and also pull attributes from the `Pulsar` class as needed. I have also made convenience functions so that we still have the old `FourierBasisGP`, `EcorrBasisGP`, and `FourierBasisCommonGP` (although the latter now takes a `spectrum` and `orf` Function instead of a cross spectrum).

I may add a few more tests but this is pretty much ready to go.


jellis18:
OK, I've added a few more tests for parameterized basis and ORFs. I think we are good to go. Anyone want to take a look before merging.

jellis18:
OK, as you can see I have had a hell of a time getting the tests to pass. There is some black magic going on with Travis and the `test_pulsar.py` test where it keeps segfaulting. For now I think it is best to drop this test and try to figure this out later. I know this is not optimal but this merge is needed for other PRs and the tests all pass on my computer...

I will be merging soon.

PR #97, opened 2017-06-22
add 'g' to backend flags

paulthebaker:
Parkes uses `-g` for backend flag in `.tim` files.


jellis18:
Ok do you want to make any other changes? If not we can merge.

paulthebaker:
This doesn't really fit in with any other changes...

PR #101, opened 2017-06-27
Fixed tests?

jellis18:
I think these changes have fixed the tests. At least they pass in my fork...


jellis18:
Ok well it appears that tests will work with `nosetest` but not with `pytest`. Right now I don't really care why this works as long as it does work. Merging.

PR #109, opened 2017-07-10
PINT compatibility

jellis18:
So here is a first stab at `PINT` compatibility. I have re-factored `pulsar.py` to have a base `BasePulsar` class that has all of the standard properties that the user calls, then I have `Tempo2Pulsar` and `PintPulsar` that inherit `BasePulsar` and get the needed information from either a `libstempo` instance or from a `TOA` and `TimingModel` instance depending on whether or not you are using `tempo2` or `PINT`.

From this I now have a *function* `Pulsar` that you initialize a `Pulsar` class from. It has the same API as the old `Pulsar` class except it now has an extra argument `timing_code`, which defaults to `tempo2`. You can also initialize this with the usual par and tim file, or you can initialize it with `libstempo` or `PINT` objects.

One reason for doing things this was is that the `Tempo2Pulsar` and `PintPulsar` classes act as a bridge between codes. Here we have a way for `enterprise` to call either code but now `libstempo` or `PINT` could call enterprise by initializing these classes internally.

I have to set up more tests but this is coming along. Also speaking of tests, I switched from using `setUp` to `setUpClass` in the unit tests since `setUp` is called *every* time a test function is called whereas `setUpClass` is called only at the beginning of each test case. This takes the test runtime down to a few minutes from what was about 20 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,8 @@ install:
     - pip install libstempo --install-option="--with-tempo2=$TEMPO2"
 
 script:
-    - coverage run --source enterprise setup.py test
+    #- coverage run --source enterprise setup.py test
+    - make test
     - make lint
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ install:
     - pip install libstempo --install-option="--with-tempo2=$TEMPO2"
 
 script:
-    #- coverage run --source enterprise setup.py test
     - make test
     - make lint
 
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -7,7 +7,6 @@ from __future__ import (absolute_import, division,
 
 import enterprise
 import numpy as np
-import tempfile
 from ephem import Ecliptic, Equatorial
 import os
 import json
@@ -17,14 +16,17 @@ try:
 except:
     import pickle
 
-from collections import OrderedDict
-
-# NOTE: PINT interface is not yet available so just import libstempo
 try:
     import libstempo as t2
 except ImportError:
     print('ERROR: Must have libstempo package installed!')
     t2 = None
+
+import pint.toa as toa
+import pint.models.model_builder as mb
+from pint.models import TimingModel
+from pint.residuals import resids
+import astropy.units as u
 
 
 def get_maxobs(timfile):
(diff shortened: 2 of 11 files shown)


jellis18:
Ok I've been doing some comparisons of `PINT` and `tempo2` just in terms of the data and the likelihood when using each code, respectively. I think the first two plots are known to the `PINT` team and it has something to do with the ephemeris or the earth barycentering method.

Different in SSB frequencies. I admit that I may be using the wrong frequencies from `PINT` (`model.barycentric_radio_freq`? sounds right...). This is a very small difference since the units are MHz but you can see the yearly trend.
pint_tempo2_freq

Difference in Residuals. Again you see the yearly trend and if this is removed I think you end up with the difference see in [this PINT issue](https://github.com/nanograv/PINT/issues/309).
pint_tempo2_res

log10 relative error in log-likelihood with parameters drawn randomly from prior. This uses the 9-year B1855+09 data set with a full noise model (efac, equad, ecorr, red noise). Overall these are pretty small but I'm a bit surprised they aren't closer. We definitely need to figure this out!
pint_tempo2_like_difference

Maybe some `PINT` people will be interested in this @scottransom.

paulthebaker:
If it's reproducing what `PINT` does, I think you should merge this.  We inherit `PINT`'s known issue.

It's really troubling that some likelihood calls have >1e-3 relative error.  That's way more than roundoff... At the IPTA meeting this came up and some people brushed it off as single v. double precision.  This looks worse than that.

jellis18:
> It's really troubling that some likelihood calls have >1e-3 relative error. That's way more than roundoff... At the IPTA meeting this came up and some people brushed it off as single v. double precision. This looks worse than that.

Thats why I don't want to merge it yet. There isn't anything obvious but I think there may be a bug somewhere. All the tests only test self consistency, in other words, they assume that the data in `Pulsar` is correct.

scottransom:
The plots are really interesting.  I suspect (as you mention) that the first two are the known issue (which we don't understand) about the tiny ephemeris differences.  I agree that fractional error differences as large as 1e-3 is worrisome, though!

jellis18:
Ok I took a look at what parameters correspond to the large difference in log-likelihood and it was for very small values of efac ~0.1. I think this makes sense because it exacerbates the differences i the residuals. In reality we will never have efacs that low. I re-ran with the lower prior range of 0.8 and I get something much more reasonable.

pint_tempo2_like_difference

I've also done a run using the tempo2 residuals with everything else PINT and get log likelihood differences never more that 1e-7 so I think this all boils down to the residuals, it just varies with noise because that can make the difference more or less pronounced.

PR #111, opened 2017-08-04
Fixed bug where Pulsar could not read parfile from current directory

jellis18:


paulthebaker:
Will the "coverage" commit to `.travis.yml` fix the greater travis issues?

jellis18:
@paulthebaker, it seems to fix problems for now. Its not really the coverage part that changes things, its that using `setup.py` for tests uses `nosetests` instead of `pytest`. For whatever reason `pytest` and `travis` do not really seem to get along. I mean they both run the same tests its just that `pytest` is neater and more clean but I don't want to spend any more time trying to debug `travis`. So for now, this allows us to still  have the tests and hopefully to not worry about random `travis` seg-faults.

PR #112, opened 2017-08-12
A few Python 3 compatibility tweaks, plus minimal vector parameter functionality

vallis:
`Parameter` factories now can be given a `size` parameter (stored as the class variable `_size`), which affects computation of priors, sampling, and display. There is no checking that the correct number of components is passed. However this minimal change allows to do a free-spectrum analysis by setting, for instance,

```
@signal_base.function
def freespectrum(f, log10_A=None, gamma=5):
    # double up the prior for cos/sin components
    # (can you do less obscurely?)
    log10_A = np.vstack((log10_A,log10_A)).T.flatten()

    # maybe this is a silly parametrization... what do we do normally?
    return ((10**log10_A)**2 / 12.0 / np.pi**2 * const.fyr**(gamma-3) * f**(-gamma))

pl = freespectrum(log10_A=parameter.Uniform(-18, -12, size=30))
rn = gp_signals.FourierBasisGP(spectrum=pl, components=30)
```

Currently these vector `Parameter`s have the same prior applied to every component. That's not a huge limitation since to do something different one would have to specify a user prior anyway.

Also care must be taking when converting vectors of sampling parameters into dictionaries. I thought we had a facility for doing that, but I can't find it anymore.


jellis18:
Can you change the way that it looks for `PINT`? At the moment it uses `pint=None` if not found but the code will fail if it does have `PINT` because you only define the `pint` variable if it isn't found. Maybe just use `have_pint` be a boolean with True if found and False otherwise. Then later in the code use `if have_pint` instead of `if pint`.

I also have a question about how we would initialize the vector parameters. It seems to me that they need to have different names to go into the parameter vector for sampling. Do you know you would handle that?

jellis18:
Do we want to add this functionality for constant parameters?

vallis:
I'm sorry, what functionality?

jellis18:
@vallis, the vector functionality. However, I think it can already handle it since it just returns a value, which can be anything.

jellis18:
OK, well I think we are good to merge then. @vallis, do you agree?

vallis:
Sure, go ahead. Once I merge into my fork I will begin experimenting with hyperparameters for priors. Unless you're doing that already...

jellis18:
I was going to take a look at hyper-parameters but I think you have a better idea of how to handle it. Merging now.

PR #115, opened 2017-08-18
General white noise class factory

jellis18:
This PR implements a generic WhiteNoise class factory that is built with a `varianceFunction` which returns the the diagonal of the N-matrix. We will still have the same public API for EFAC and EQUAD as I've subclassed the generic class, but now we could add something like DMEQUAD with

```python
# define DM EQUAD variance function
@function
def dmequad_ndiag(freqs, log10_dmequad=-8):
    return np.ones_like(freqs) * (1400/freqs)**2 * 10**(2*log10_dmequad)

# return DM EQUAD factory
variance = dmequad_ndiag(log10_dmequad=parameter.Uniform(-10, -5))
dmeq = WhiteNoise(variance)
```

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -423,7 +423,7 @@ def Pulsar(*args, **kwargs):
         elif timing_package.lower() == 'tempo2':
 
             # hack to set maxobs
-            maxobs = get_maxobs(reltimfile)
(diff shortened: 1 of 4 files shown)


jellis18:
Ok, these are fairly minimal changes but could be really useful for people who want to define new white noise signals. I'm going to merge now.

PR #117, opened 2017-08-25
Python 3 compatibility: there's no cPickle, and pickle files need to be explicitly opened as binary

vallis:


jellis18:
Ok, I missed the binary part when reading the files back in when I played around with this. I have a branch that has some other python3 fixes and adds python 3.5 and 3.6 to the travis tests. I'll incorporate your changes there and make a python3 compatibility PR.

PR #119, opened 2017-08-25
Preliminary support for parameter priors with hyperparameters

vallis:
So this is my first attempt. I think it breaks some of the `Signal` functionality, but I wanted your feedback before proceeding. The idea is to create a new type of parameter, `UserParameter`, which takes an Enterprise `Function` as the prior. The `Function` is instantiated when the `Parameter` is, so all the `Function` parameters get specialized names.

This means also that `get_pdf` and `get_logpdf` must take parameter dictionaries, so they can compute the prior appropriately. Last, now `Parameter` has its own `params` method, which returns a list containing the `Parameter` instance itself, plus any hyperparameters that affect the prior. The `Signal` classes that accept parameters only through functions (such as the generic white-noise and generic GP signals) are already fine; the other `Signal` classes probably need fixing to iterate correctly over all the parameters.

With these changes I can do

```python
>>> import enterprise.signals.parameter as esp
>>> import enterprise.signals.signal_base as ebb

>>> @esb.function
>>> def gaussianprior(x, x0=0, std=1):
>>>     return np.exp(-0.5 * (x - x0)**2 / std**2) / np.sqrt(2 * math.pi * std**2)

>>> b = esp.UserParameter(gaussianprior(x0=esp.Uniform(-1,1), std=esp.Uniform(-2,2)))

>>> b1 = b('gauss')

>>> b1
"gauss":UserParameter(gauss_std,gauss_x0)

>>> b1.params
["gauss":UserParameter(gauss_std,gauss_x0),
 "gauss_std":Uniform(-2,2),
 "gauss_x0":Uniform(-1,1)]

>>> b1.get_pdf(0.5) # uses defaults
0.35206532676429952

>>> b1.get_pdf(params={'gauss': 0.5, 'gauss_x0': 1, 'gauss_std': 2})
0.19333405840142465
```

and also

```python
>>> def f(x, a=0):
>>>     return x + a

>>> c = esb.Function(f,a = esp.UserParameter(gaussianprior(x0=esp.Uniform(-1,1), std=esp.Uniform(-2,2))))

>>> c1 = c('secondary')

>>> c1.params
["secondary_a":UserParameter(secondary_a_std,secondary_a_x0),
 "secondary_a_std":Uniform(-2,2),
 "secondary_a_x0":Uniform(-1,1)]

>>> c1(0,params={'secondary_a': 2})
2
```

Comments welcome!


jellis18:
I think this is a good start but I think it would be better if maybe we modify the existing parameter class to recognize `Parameter`s as arguments. That way we could just do
```python
p1 = Normal(Uniform(-1, 1), Uniform(1, 10)) 
```

We could still have `UserParameter` for special user defined cases. For instance, we still need a way to have the parameter get info from the `Pulsar` class as we will need for pulsar distances and timing parameters.

jellis18:
Is there a reason you copied the `Function` definition from `signal_base`?

vallis:
Regarding moving `Function` to `parameter.py`, it's because I need it in `parameter.py` to handle hierarchical priors, but `signal_base.py` imports `parameter.py`, so I had to resolve a cycle of imports.

jellis18:
Are these changes finished?

vallis:
I think I may be done for now... the `__repr__` of `Parameters` and `Functions` may not be ideal yet, and parameter groups, as well as repeat parameters, remain to be figured out. But hierarchical priors work, with the caveat that I discuss in https://github.com/nanograv/enterprise/pull/119/commits/2a6abe737e44d21bbdfa1cbbd8cae51872c158da

vallis:
I'm going to close this PR, which at this point is too old, and open a new one with fresher code.

PR #125, opened 2017-09-26
Physical ephemeris model

jellis18:
I have implemented the physical ephemeris model that Michele and Steve have been working on. It all seems to work and I can re-produce the 11-year results but I'd still like to do a direct check to compare `enterprise` with NX01 waveform.

I have implemented this through the "waveform function" in `utils.py` called `physical_ephem_delay`. This is an `enterprise`  `Function` and will take in the physical parameters and output the delay in seconds (this already has the roemer delay subtracted off). 

To implement the class factory I have defined `PhysicalEphemerisSignal` which has the default priors that Steve has used but these can easily be changed by the user. The factory then initializes the waveform with the parameters and feeds it to the generic `Deterministic` class factory which is subclassed in `PhysicalEphemerisSignal`. We have to do some initialization but overall it isn't that bad.

The API looks like
```python

# set up with default parameters
# this has default parameter priors, includes jupiter orbital modeling, no saturn modeling
# and uses epoch TOAs to speed up the calculation
eph = deterministic_signals.PhysicalEphemerisSignal()

# to change the boolean parameters you can do something like 
eph = deterministic_signals.PhysicalEphemerisSignal(inc_jupiter_orb=True,
                                                    inc_saturn_orb=True, 
                                                    use_epoch_toas=False)

# or you can pre-define andy or all of the parameters before hand 
# if you want to use different priors

# drift rate
drift = parameter.Uniform(-1e-9, 1e-9)('frame_drift_rate')

# mass perturbation parameters
dj = parameter.Normal(0, 1.54976690e-11)('d_jupiter_mass')
ds = parameter.Normal(0, 8.17306184e-12)('d_saturn_mass')
du = parameter.Normal(0, 5.71923361e-11)('d_uranus_mass')
dn = parameter.Normal(0, 7.96103855e-11)('d_neptune_mass')

# jupiter orbital elements
joe = parameter.Uniform(-0.05, 0.05, size=6)('jup_orb_elements')

# saturn orbital elements
soe = parameter.Uniform(-0.5, 0.5, size=6)('sat_orb_elements')

eph = deterministic_signals.PhysicalEphemerisSignal(
    frame_drift_rate=drift, d_jupiter_mass=dj, d_saturn_mass=ds, 
    d_uranus_mass=du, d_neptune_dn, jup_orb_elements=joe, 
    sat_orb_elements=soe, inc_jupiter_orb=True, 
    inc_saturn_orb=True, use_epoch_toas=True)
```

I have also written some tests that just check to see if we are calling the `physical_ephem_delay` function correctly and to see if the epoch-TOA waveform is the same as the full TOA waveform.

Let me know what you all think.


vallis:
This looks fine, Justin. Thanks for writing it.

I owe you a jointly normal vector Parameter so you can use our PCA prior.

stevertaylor:
Looks great Justin. I really like the API. I'm keeping many plates spinning at the moment, but I'll try to get you a comparison waveform soon.

jellis18:
OK, so @stevertaylor and I have compared waveforms and they now agree. I did have a bug previously in that I forgot to convert `psr.planetssb` and `psr.pos_t` from ecliptic to equatorial coordinates.

Issue #122, opened 2017-09-20
Speed up basis matrix column checking


jellis18:
Currently we always go through every column of every basis matrix to check if any columns are identical. This is very useful since it allows us to use an arbitrary number (and value) of frequencies for different red noise processes without using more columns in the overall basis matrix than we need. For example, if we are using the same 30 frequencies for red noise and  GWB then the overall basis matrix should only have 60 values (sine and cosine double the value) and not 120. 

This all works fine but the method that we are currently using is very brute force and **very** slow. It takes several minutes for the newer data sets. This is not terrible since we only have to do it once but it is still a bit annoying.

Does anyone have any thoughts on how to speed this up?

vallis:
Assuming that the columns would be exactly identical (not just very close), some form of hashing could speed up the comparison. Is that really what's most expensive though, or is it the repeated computation of the same column?

jellis18:
Its the repeated computation. One way to avoid this (which brings up another issue) is to only do this for the Fourier basis components, since those are the only ones that would need this functionality anyway.

The problem with this is that currently there is no enforced naming of signals. All signals have a `signal_type` and `signal_name` but there is nothing that requires that these are set of that these are different for each signal or that they even bare any resemblance to the signal itself.

Maybe we should nave a name required for each signal that would then be prepended with the pulsar name when initialized...

jellis18:
> The way we're doing them now, the comparisons seem very wasteful. Are they not a factor in timings

Yes, maybe we are misunderstanding each other. They are a huge factor. The problem is that the overall basis matrix can be up to 40,000 x 1,000 and we are doing comparisons of every row with every other so it goes like 1e6 comparisons of 40,000 dimensional vectors. The large number of columns comes from the timing model and/or ECORR if done in the basis way. If we only had the fourier basis then it wold still be inefficient but we would only need to compare a few 100 columns at most

vallis:
Ah then a hash is the way to go. Sounds fun, I will try

vallis:
Something like this should work...

```python
        def _combine_basis_columns(self, signals):
            """Given a set of Signal objects, each of which may return an
            Fmat (through get_basis()), return a dict (indexed by signal)
            of integer arrays that map individual Fmat columns to the
            combined Fmat.

            Note: The Fmat returned here is simply meant to initialize the
            matrix to save computations when calling `get_basis` later.
            """

            idx, Fmatlist, hashlist = {}, [], []
            cc = 0
            for signal in signals:
[...]

PR #126, opened 2017-10-05
Optimizing basis column lookup

jellis18:
This fixes Issue #122.


jellis18:
Ok, I may have a solution for this. Since the problem with the fourier columns come from very small differences in the Fourier frequencies due to using `np.linspace`. We could use `np.arange` but as @vallis mentions in #122, this only works for the standard frequency spacing and not more general spacings. 

So I have just done this at the end of creating the frequencies `f`, `f = np.double(f.astype('float32'))`. This does seem to fix the problem as well but seems a little sketchy.

What do you think?

Issue #30, opened 2017-03-08
Parameterization and units of PSD (Fourier design matrix)


vhaasteren:
I'd like to raise an issue that has been bothering me in all our previous codes for a while: we have always been using the incorrect units for the PSD, as first defined in Lentati et al. (2013). Enterprise is the place to do it properly. However, there are some caveats that have to do with numerical stability, so let me explain.

The function createfourierdesignmatrix_red contains the relevant code. We always create the Fourier Design Matrix (F) as consisting of basis functions that are essentially of the form cos(x), so with a range of [-1,1]. Doing this places the binsize of the PSD (1/T) in the definition of the PSD (diagonal of phi), which makes the units of phi sec^-2. The actual PSD is of course independent of the binning, and should have units of sec^-3. I would propose the following:
Place the binsize in the definition of F, so that the PSD becomes independent of the binning, and has proper units.

The caveats I can see are the following:
- the mode parameters ('a', the ones that one typically analytically marginalizes over), now no longer have the units of sec, which makes them less well-interpretable. However, since they pretty much always have a modeled prior on them, I don't think this is an issue. When sampling, this is therefore not really an issue.
- When using complicated PSD parameterizations with different sized binning at different frequencies (like in van Haasteren & Vallisneri, 2014), the matrix F no longer scales with the original F (with 1/T included), since different columns are scaled differently. However, I don't see a real reason for this to cause numerical instability.
- In a way, the binning is actually a model parameter, which makes F dependent on a model parameter. However, covertly that is already the case in the current function, since the binning also requires one to set the PSD sampling frequencies, which _are_ already passed to createfourierdesignmatrix.
- The results of the package would have to be scaled to compare with previous results. However, in my memory, there aren't many significant results that have been presented in papers that use the units of sec^-2. Those are mostly just internal values.

stevertaylor:
I know it's been a long delay on this, but I just wanted to say that I agree. I'm in the process of changing my code to incorporate the summing weights into the design matrix definition. This became necessary when I wanted to start using your log-spaced frequencies.

PR #132, opened 2017-11-01
2D Kernels

jellis18:
This PR implements 2-D GP kernels which solves some problems mentioned in Issue #93. Previously all gaussian process kernels (or priors, think diagonal phi matrix) were assumed to be 1-d (i.e. diagonal). These new commits don't change the API at all so all of the old GP modeling stuff still works the same, but you can now do something like the following to implement a Squared-Exponential Kernel in the daily averaged basis (averaging over 7 days)
```python
@signal_base.function
def create_quant_matrix(toas, dt=1):

    U, _ = utils.create_quantization_matrix(toas, dt=dt, nmin=1)
    avetoas = np.array([toas[idx.astype(bool)].mean() for idx in U.T])
    return U, avetoas

@signal_base.function
def se_kernel(avetoas, log10_sigma=-7, log10_lam=np.log10(30*86400)):
    tm = np.abs(avetoas[None, :] - avetoas[:, None])

    # add to diagonal to make stable
    d = np.eye(tm.shape[0]) * 10**(2*(log10_sigma-1.5))
    return 10**(2*log10_sigma) * np.exp(-tm**2/2/10**(2*log10_lam)) + d

# define signal
log10_sigma = parameter.Uniform(-10, -5)
log10_lam = parameter.Uniform(np.log10(86400), np.log10(1500*86400))
basis = create_quant_matrix(dt=7*86400)
prior = se_kernel(log10_sigma=log10_sigma, log10_lam=log10_lam)
se = gp_signals.BasisGP(prior, basis, name='se')
```

I have written new tests for this and it all seems to work fine with and without spatial correlations except when using the "cliques" method for spatial correlations. @vallis, can you maybe take a look at this?

One other thing that I changed here is how the PSDs are handled which I think clears up Issue #30. Before the "prior" for PSDs, `powerlaw` for instance, would return something in s^3 but the Kernel is expected to be a variance. We would get around this by multiplying by f[0] (same as 1/Tspan) when we constructed the actual variance but this is not general because it wouldn't work for anything but linear frequency spacing and also I had to use this same formalism for all other Basis type signals which means I was multiplying the variance (ECORR) for instance with ones. 

So the new formalism assumes that the "priorFunction in `BasisGP` actually returns a variance. For all other signals this was already the case but now I have changed the PSD definitions to include the df piece so now when they read in the frequency, f, it computes df by `df=np.diff(np.concatenate(np.array([0]), f))`. This way it is correct for whatever type of frequency spacing we have!

Also, as you can see form the commit tags, I'm having problems with Travis again :(.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,8 @@ install:
     - cd ../
 
 script:
-    - coverage run --source enterprise setup.py test
+    - make test
+    #- coverage run --source enterprise setup.py test
     - make lint
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 #cache: pip
-#dist: precise
+dist: precise
 
 python:
     - "2.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,39 @@ install:
     - cd ..
     - rm -rf psrsoft-tempo2-*
 
+    - sudo apt-get update
+    # We do this conditionally because it saves us some downloading if the
+    # version is the same.
+    - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
+        wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh;
+      else
+        wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+      fi
+    - bash miniconda.sh -b -p $HOME/miniconda
+    - export PATH="$HOME/miniconda/bin:$PATH"
+    - hash -r
+    - conda config --set always_yes yes --set changeps1 no
+    - conda update -q conda
+    # Useful for debugging any issues with conda
+    - conda info -a
+
+    # Replace dep1 dep2 ... with your dependencies
+    - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION 
+    - source activate test-environment
     - pip install -r requirements_dev.txt
     - pip install -r requirements.txt
     - git clone https://github.com/vallis/libstempo.git
     - cd libstempo
     - python setup.py install --with-tempo2=$TEMPO2
     - cd ../
+    - python setup.py install
+
+    #- pip install -r requirements_dev.txt
+    #- pip install -r requirements.txt
+    #- git clone https://github.com/vallis/libstempo.git
+    #- cd libstempo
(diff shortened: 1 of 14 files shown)


jellis18:
One other thing that we probably want to do is to change the way the Basis functions are defined. As of now all basis functions return the basis matrix and a parameter that may or may not be used in the prior Kernel. For example, the Fourier basis returns the F matrix and the sampling frequencies. The PSD type priors then read in the sampling frequencies here. The Kernel above returns the quantization matrix and the average TOAs, which are then used in the Kernel. However, what if we need more than one variable.

I think we should change the output of Basis functions to be the basis matrix and a dictionary of parameters, so the fourier design matrix would return `return F, {'f': f}` and then the PSD can be called internally like `powerlaw(params=params, **labels)`, where `labels` are the dictionary.

This will require re-writing some of the tests but I think it is worth it, especially once we try to model things in time and radio frequency we are going to need both average TOAs and average frequencies.

vallis:
Thanks, Justin. Comments:

- I will look at the cliques tomorrow, but perhaps in the meantime you can accept the request. Are GPs usable without cliques?
- Regarding PSD integration, if the PSD is evaluated at f_0, f_1, f_2, f_3, then it would be more accurate for f_i to multiply 0.5*(f_{i+1} - f_{i-1}), and for f_0 and f_N to multiply half a step. Or even to use Simpson's rule (see Rutger's and my http://www.arxiv.org/abs/1407.6710). That's not what we do now, but it shouldn't affect limits/Bayes ratios.
- Regarding passing the "labels" to powerlaw functions, I think it works (at least in my PR) to pass positional arguments in "params"---that is, powerlaw(f, t, ...) should be able to take everything from params.

jellis18:
Ok so I have to figure out the travis stuff but there are at least two more things I want to do with this.

1. Add tests to make sure 2-D kernels work with selection
2. Modify so that the inverse method will only do cholesky if the array is 2-D. This is important in `SignalCollection` as right now it forms one phi matrix (if any signals have 2-D) and inverts

jellis18:
>>1. Add tests to make sure 2-D kernels work with selection

I actually already did this!

>>2. Modify so that the inverse method will only do cholesky if the array is 2-D. This is important in SignalCollection as right now it forms one phi matrix (if any signals have 2-D) and inverts

This would speed things but may be too much work for this PR. I think it would require a big re-write of KernelMatrix.

@vallis, what do you think?

jellis18:
OK, I'm going to merge this now, the optimization in point 2 above can be taken care of in a different PR. This PR implements the functionality and is tested so lets do it!

PR #157, opened 2018-03-19
Added orfs for anisotropy

stevertaylor:
Added anisotropy ORFs. Alternative polarization ORFs to follow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
+.ipynb_checkpoints/
 
 # PyBuilder
 target/
--- a/docs/_static/notebooks/.ipynb_checkpoints/data-checkpoint.ipynb
+++ b/docs/_static/notebooks/.ipynb_checkpoints/data-checkpoint.ipynb
@@ -1,0 +1,747 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# `enterprise` Data Structures\n",
+    "\n",
+    "This guide will give an introduction to the unique data structures used in `enterprise`. These are all designed with the goal of making this code as user-friendly as possible, both for the end user and the developer."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2017-12-08T01:37:29.763231Z",
+     "start_time": "2017-12-08T01:37:29.748500Z"
+    }
+   },
+   "outputs": [],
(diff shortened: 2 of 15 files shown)


paulthebaker:
Hi Steve, I've look these over a bit.

* The `lint` tests are failing.  you should run `make lint` locally to see the errors.
* The unit tests failed.  I think because `test_orf` needs to take `self` as an argument.
* I can help fix these

* I'm adding `.ipynb_checkpoints/` to `.gitignore` on master.  There's no reason to keep those in the repo.  When that PR passes checks and merges I'll update this from `master`.

stevertaylor:
Thanks Paul. I guess the notebook checkpoints were added automatically when I did a git add. Thanks for adding them to gitignore. Surprised we didn't have them in there already. 

I'll modify the tests now.

paulthebaker:
I opened a PR to your branch with lint fixes.  The pytests run, but `test_orf` still fails:

```
AssertionError: ORF auto term incorrect for anisotropy
assert 2.2309999999999999 == 1.147
```

stevertaylor:
I was in the process of fixing all the lint and tests myself. The test runs fine on my machine.

paulthebaker:
On a potentially annoying API level `anis_coefficients.anis_basis` takes a list pulsar positions given as (theta, phi), where all of the `utils.xxx_orf` functions take pulsar positions as a unit vector.  When you get deep enough into `anis_basis` it looks like the unit vectors are made anyways.  Why not just pass the top level function the unit vectors?

stevertaylor:
What was in the changes you committed to this PR, Paul?

paulthebaker:
https://github.com/stevertaylor/enterprise/pull/1

stevertaylor:
Thanks for doing this. I also made these changes, but I guess now need to write over them when pulling. Is test_orf still failing for you?

stevertaylor:
For the API issue, I feel like the phi, theta coordinates rather than the unit vectors are the easier and more useful quantities to deal with. The unit vectors make sense when you're only dealing with angular separations between pulsars, but we want more general ORFs. I agree that in some places in anis_coefficients you are making the unit vectors anyway, but not in all places. But I could switch over to unit vectors in a subsequent PR if needed.

paulthebaker:
Re: API, most of this is under the hood, so it's not a big deal.

stevertaylor:
Where can I find information on why this is failing?

paulthebaker:
The logs are on the travis site: https://travis-ci.org/nanograv/enterprise.

Most recent log: https://travis-ci.org/nanograv/enterprise/jobs/356002583.

Currently, it looks like we need to add `healpy` to the requirements, so travis installs it before running the tests.

PR #159, opened 2018-04-08
Added option to declare BIPM version in Pulsar object

stevertaylor:


stevertaylor:
The build is failing because enterprise is testing against an older version of pint that does not have bipm_version as an option to pint.toa.get_TOAs.

stevertaylor:
The tests should pass if we updated the PINT tag in the requirements.txt file to a newer version that includes a bipm_version argument in pint.toa.get_TOAs. Any problem with me doing this?

paulthebaker:
Lets update to the latest PINT.  If that breaks something else, we can sort it out.

Issue #164, opened 2018-07-03
Normalization for DM in utils.createfourierdesignmatrix_dm 


bvgoncharov:
Hi All,

I have recently started to use Enterprise, and while resolving a discrepancy of my results with other person results (from PAL2) noticed the following. In [utils.createfourierdesignmatrix_dm](https://github.com/nanograv/enterprise/blob/2f7c3f48d5feac809819696e98ed095666bdcc55/enterprise/signals/utils.py#L107) we create a Fourier design matrix for DM variation noise. But it looks like there is a to-do item: # TODO: should we use a different normalization. 

Here are my suggestions:
1) Replace [line 134](https://github.com/nanograv/enterprise/blob/2f7c3f48d5feac809819696e98ed095666bdcc55/enterprise/signals/utils.py#L134) with the following:
`Dm = 4.15e3/(freqs**2)`
2) Add a comment:
`# 4.15e3 = K^(-1), where K is from Eq. 23 in L. Lentati paper: https://academic.oup.com/mnras/article/437/3/3004/2907742`

Do you think these are worth implementing?
I would be happy to git-push these changes.

jellis18:
I would rather keep the current normalization which makes the basis functions dimensionless. There are several reasons for this.

1. Much easier to gauge "strength" of DM signal as with this normalization it is simply the deviation (in seconds) at 1.4 MHz.
2. Allows the exact same parameterization and priors as we use for red noise.
3. Keeps code uniform in that all basis functions are dimensionless and all gaussian priors are variances in units of seconds^2.
4. Allows for easy extension to other exponents on radio frequency (i.e. scattering, or varying, etc) without having to figure out what other constant to use for normalization

Of course this is up to the other developers as well but I strongly prefer the dimensionless normalizations.

paulthebaker:
I also prefer to keep them dimensionless, but it would be pretty easy to allow the `create designmatrix` functions to take a normalization option.  The hardcoding of numbers always bothers me, even if I would choose 1400 MHz too.

bvgoncharov:
@jellis18 @paulthebaker then we could just add "fref" argument in utils.createfourierdesignmatrix_dm, with a default of 1400, and mention that it has MHz units in the comment. It shouldn't affect existing users

jellis18:
@bvgoncharov, that sounds good. Do you want to implement this and make a PR?

bvgoncharov:
[...]

PR #165, opened 2018-07-16
Introducing an option to choose a reference frequency for DM normalization

bvgoncharov:
This pull request is related to the following topic: [https://github.com/nanograv/enterprise/issues/164](https://github.com/nanograv/enterprise/issues/164)


paulthebaker:
We aren't using camelcase for any other of the kwargs (`fmax` not `fMax`).  Can you rename the new parameter `fref` to match this style.

Can you move your comments to the docstring e.g. `:param fref:`

We haven't linked to github issues in comments before, so you should remove that too.  @jellis18, do you agree?

bvgoncharov:
@paulthebaker ok, let's wait for @jellis18 to confirm.
As for the issue link, I just thought it would be nice to have some references to default choices of models and normalisations for Enterprise. Perhaps we could add this information to https://enterprise.readthedocs.io/ ? (with or without a link to the issue page) I would be happy to write a short note about DM and red noise normalisation and models for readthedocs.io.

jellis18:
Yes I would follow @paulthebaker's instructions. The explanation and default choice should be in the doctoring `param: fref` but there shouldn't be a link.

paulthebaker:
The docstrings should be used to create autogenerated docs in the module index.  We should probable link that in the sidebar (in addition to the link at the very bottom of the landing page).

Looking now, I see that in the module index doesn't have any entries for `signals.utils` or `signals.white_signals`.  I'll look into why when I get a chance.

bvgoncharov:
@jellis18 @paulthebaker ok, sounds good. I removed the link and moved a comment to the docstring. The comment is quite general, so I moved it to the top, near the description. But please let me know if you think it's more appropriate near `param: fref`

jellis18:
Looks good except `fref` is in MHz not KHz.

bvgoncharov:
@jellis18 , I fixed this. Sorry, I suspected that everything should be in MHz, but got confused by your post [on the issue page](https://github.com/nanograv/enterprise/issues/164):
`Much easier to gauge "strength" of DM signal as with this normalization it is simply the deviation (in seconds) at 1.4 MHz.`
Is must be 1.4 GHz then.

paulthebaker:
Why isn't this letting me squash and merge?  There seems to be a pending status check, but I can't figure out what it is...

bvgoncharov:
@paulthebaker I am not sure. The messages I see below your comment are: "All checks have passed" and "This pull request can be automatically merged by project collaborators"

PR #166, opened 2018-07-24
Hierarchical Functions and Parameters

vallis:
This PR, many months in gestation, allows Functions to take other functions as Parameters (tracking parameter dependencies correctly), and as an application it allows Parameters to have priors that depend on hyperparameters.

Tests pass and I have written simple tests for the hierarchical functionality.

(The PR includes also minor changes to pulsar.py to represent libstempo DM parameters. It was easier to include these than to separate them into a different commit.)


paulthebaker:
I have a few comments on styling... sorry.  I can make these changes and make a PR into your PR if you want.

Standardizing imports
----------------------

There are a few imports that are not standardized across different files.  In `gp_signals.py`, `white_signals.py`, `test_deterministic_signals.py`, and `test_vector_parameter.py` you import `parameter.function` as `enterprise_function`.  Whereas in `signal_base.py` and `utils.py` you simply import `function` without a rename.

For readability, we should have a standard import statement.  Either 
```python
from enterprise.signals.parameter import function as enterprise_function
```
or
```python
from enterprise.signals.parameter import function
```

Docstrings
-----------
Can we move usage comments into the docstrings of class factories functions.  This will put these comments in the auto-generated docs.  While we're at it we should put `:param ???:` and `:returns:` tags in the class factory docstrings  For example:

```python
def Uniform(pmin, pmax, size=None):
    """Class factory for Uniform parameters.
    This handles vectors correctly, if ``pmin`` and ``pmax`` are scalars,
    or if ``len(size) == len(pmin) == len(pmax)``

    :param pmin: minimum of uniform range
    :param pmax: maximum of uniform range
    :return: ``Uniform`` parameter class
    """

    class Uniform(Parameter):
        _size = size
        _prior = Function(UniformPrior, pmin=pmin, pmax=pmax)
        _sampler = staticmethod(UniformSampler)
        _typename = _argrepr('Uniform', pmin=pmin, pmax=pmax)

    return Uniform
```

vallis:
Thanks Paul, very useful. I've changed what I could find immediately...

Since you have a great eye for these, you shouldn't be shy about making such changes throughout the code whenever needed (and even short-circuit PRs, I won't complain).

PS You can make a PR for a PR? :)

vallis:
Hello Paul, do you have a Python 3.5 installation to diagnose the test that's failing? I have 2.7 and 3.6 and would rather not install another one... Thanks!

paulthebaker:
I also use 2.7 and 3.6.  I will whip up an anaconda virtual env this morning for it.  Sorry, I didn't test 3.5 before saying it was ready to go...

paulthebaker:
Running `make test` in my virtual env fails because I can't get `healpy` to build.  According to the Travis logs, the failures were all in `tests/test_hierarchical_parameter.py`.   Simply running `py.test` for that file passes.  I restarted the build on travis to see if it works this time.  I'll work on getting `healpy` up in the meantime.

paulthebaker:
So I got `healpy` to build.  A local run of `make test` and the exact `coverage ...` command travis is running both work.

My virtual env had some version differences from the `requirements.txt`.  I down graded to `numpy==1.13.3` and got the errors.

```
>       assert str(x1) == 'x1:Uniform(pmin=0, pmax=1)'
E       AssertionError: assert 'x1:Uniform(pmax=1, pmin=0)' == 'x1:Uniform(pmin=0, pmax=1)'
E         - x1:Uniform(pmax=1, pmin=0)
E         + x1:Uniform(pmin=0, pmax=1)

tests/test_hierarchical_parameter.py:42: AssertionError
```

All 5 of them appear to be problems in the ordering of `pmin` and `pmax`.  `numpy==1.14.5` does not have this error.  I'm guessing something with `OrderedDict` handling in `kwargs`.

@vallis, the simplest fix seems to be changing the numpy version in `requirements.txt`.

The `einsum` issue was fixed in [1.14.1](https://github.com/numpy/numpy/blob/master/doc/release/1.14.1-notes.rst).  I'm currently trying to figure out what the minimum workable version is.

paulthebaker:
`numpy==1.14.1` does not work for `test_hierachical_parameter.py`.
`numpy==1.14.2` does.  I'm now running all tests with this version...

paulthebaker:
I *think* changing `requirements.txt` to `numpy>=1.14.5` will fix it.

While we're at it, we can move to `scipy>=1.1.0`

vallis:
I'm going to pull the plug and accept this PR. All tests pass.

Note that you need https://github.com/stevertaylor/enterprise_extensions/pull/6 or https://github.com/vallis/enterprise_extensions to use exterprise_extensions with this commit.

vallis:
It is! I have another big PR coming (support explicit GP coefficients), but I'll try to keep the merge Hobbit-sized.

PR #168, opened 2018-09-26
Handle explicit GP coefficients

vallis:
This PR enables creating `BasisGP(..., coefficients=True)` objects that treat Gaussian process coefficients explicitly. The `coefficients` option can be given also to the specialized `FourierBasisGP`, `TimingModel`, `EcorrBasisModel`.

That is, the values of the coefficients must be provided as a numpy array among the parameters (with autogenerated names like `B1855+09_red_noise_coefficients`), and the resulting TOA delays are returned by `Signal.get_delay()` instead of (implicitly) `Signal.get_basis()`. The vector coefficient parameters are generated as part of `Signal` initialization, and are included in `Signal.params`; they carry the correct GP prior derived from the prior furnished to BasisGP.

Explicit GP coefficients allow sampling the hierarchical PTA likelihood (as useful, e.g., in cross-validation schemes).

Comments:

- Currently coefficients are not supported for common GPs (such as GWs).
- Currently GP coefficients cannot be sampled consistently with `parameter.sample()`. However the function `utils.get_coefficients(pta, params)` samples all coefficient vectors from the conditional distribution induced by their GP hyperparameters.
- To harmonize the naming of parameters, all GPs now carry a default name (such as `red_noise`); thus the parameter that may have been called `B1855+09_log10_A` is now `B1855+09_red_noise_log10_A`).


stevertaylor:
The checks are failing here because of a bug in the most recent version of numpy. https://github.com/gwastro/pycbc/issues/2402

stevertaylor:
We can patch this by restricting numpy to versions before the most recent: https://github.com/gwastro/pycbc/commit/17fa90d1e5b251a86caff29bc4bc302fe35a667a

stevertaylor:
A simple change to `numpy>=1.14.1,<1.15.3` would suffice to patch this until a newer numpy version plays well with astropy. We can then merge these changes.

vallis:
Thanks @stevertaylor. Applying your workaround now.

PR #172, opened 2019-02-05
Adding option (combine=False) to keep GP bases uncombined when multiple GP signals use the same Fourier modes

vallis:


vallis:
This is useful when drawing GP coefficients from predictive conditional posteriors with utils.get_coefficients(). It does not yet work with CommonGP signals (e.g., GWs with H-D correlations)

PR #173, opened 2019-02-05
Allow coefficients=True and combine=False in BasisCommonGP

vallis:


vallis:
Note: commit 1d3e4ec has also the effect of making 'cliques' (instead of 'partition') the standard method for get_phiinv(). 'cliques' is tested and it's more general, and it allows common processes with dense prior matrices. Nevertheless, proceed with caution...

PR #175, opened 2019-02-20
changed inspect.getargspec() to inspect.getfullargspec()

svigeland:
I made this change because inspect.getargspec() is deprecated in Python 3.


vallis:
Thanks Sarah! I see the tests pass, that's great. To allay any residual concerns, were you able to determine that getargspec().args and getfullargspec().args have the same semantics? (And the same for .defaults?)

svigeland:
I will check this and get back to you. I am at Ole Miss right now, and I am about to go out to dinner with Leo and Jenny, so I probably won’t respond until tonight or tomorrow.

svigeland:
> To allay any residual concerns, were you able to determine that getargspec().args and getfullargspec().args have the same semantics? (And the same for .defaults?)

According to the documentation, they seem to be the same. I ran getargspec() and getfullargspec() on a few enterprise functions, and I got the same results for getfullargspec().args and getfullargspec().defaults.

I realized, though, that inspect in python2 does not have getfullargspec(), so I added a try...except... statement so that it will use whichever one is appropriate for the user's version of python.

svigeland:
I've confirmed that inspect.getargspec().args and inspect.getfullargspec().args give the same results for all of the functions in selections.py and parameter.py. I've also checked that inspect.getargspec().defaults and inspect.getfullargspec().defaults give the same results. I think this is ready to be merged.

Issue #183, opened 2019-05-16
Change Pulsar to run if user has PINT, but not libstempo


Hazboun6:
Currently the user gets a printed message that says 'ERROR: Must have libstempo package installed!' if the import call for `libstempo` raises an `ImportError` exception. There are a number of user types who do need libstempo installed, including those dealing with gamma-ray or x-ray timing data, since libstempo (really TEMPO2) can't deal with these types of TOAs. 

It looks like the only change to the working code would need to be [here](https://github.com/nanograv/enterprise/blob/master/enterprise/pulsar.py#L453-L457) and would just be changing:
```Python
    if pint:
        toas = list(filter(lambda x: isinstance(x, toa.TOAs), args))
        model = list(filter(lambda x: isinstance(x, TimingModel), args))

    t2pulsar = list(filter(lambda x: isinstance(x, t2.tempopulsar), args))
```

to

```Python
    if pint:
        toas = list(filter(lambda x: isinstance(x, toa.TOAs), args))
        model = list(filter(lambda x: isinstance(x, TimingModel), args))
    if tempo2:
        t2pulsar = list(filter(lambda x: isinstance(x, t2.tempopulsar), args))
```

Besides that we'd just add a truthful error message. Something like: `raise ValueError('Must have either PINT or libstempo installed.')`

Does anyone know of other places in the code that might _require_ libstempo/TEMPO2?

paulthebaker:
This is a good idea.  You should also change the import behavior to give warnings if either is missing.

I don't think `PINT` or `libstempo` is used outside of loading `Pulsar` objects.

PR #184, opened 2019-05-16
Shielded libstempo dependencies if only pint installed

Hazboun6:
This is Pull Request to deal with Issue #183. I've changed the code in `enterprise.Pulsar()` so that `libstempo` is no longer required if user has`PINT` installed.

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -39,6 +39,7 @@ if pint is None and t2 is None:
     err_msg = 'Must have either PINT or libstempo timing package installed'
     raise ImportError(err_msg)
 
+
 def get_maxobs(timfile):
     """Utility function to return number of lines in tim file.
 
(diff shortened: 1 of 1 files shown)


paulthebaker:
While you're at it you could replace the `list(filter( ...))` with list comprehensions.   I think this is equivalent, but I haven't tested:

```python
[isinstance(x, t2.tempopulsar) for x in args]
```

Hazboun6:
I think what you have written will return booleans, but the `list(filter())` returns an object.

paulthebaker:
Okay.  To get a list of objects, you want:
```python
[x for x in args if isinstance(x, t2.tempopulsar)]
```

I sort of tested it... This gives me a list of `int`s:
```python
args = [0.0, 1, 2, 3.0, 4, 5.0]
[x for x in args if isinstance(x, int)]
```

Hazboun6:
As @paulthebaker suggested elsewhere, there are many instances of `list(filter())` throughout the code. This may be taken up at a later time.

PR #192, opened 2019-05-17
used logger for warning messages in pulsar.py

svigeland:
This allows you to turn off the warning messages using logger.


paulthebaker:
I think if you move `import logging` to the main import block at the top, it will fix the remaining `lint` issue.

paulthebaker:
Only the `import` statements go at the top.  other calls to `logging` need to be below the rest of the imports.

PR #201, opened 2019-06-19
Added selection functions for *single*  frequency bands and frequency-based (instead of flag-based) selection

danielreardon:
This is added so that the "band noise" selection of the IPTA dr1 noise paper (Lentati et al. 2016) can be  reproduced with enterprise. This was tested using modified functions in enterprise_extensions


paulthebaker:
I'm not super keen on hard coding in the frequency ranges or having 4 functions to do very similar things.  The by frequency function could be generalized by doing something like this:
```
def by_frequency(freqs, fsplit=[1000, 2000]):
    """Selection function to split by frequency bands
    
    :param f_splits: list of frequencies to split on (MHz)
    """
    fsplit = np.concatenate([[0], np.array(fsplit, dtype=float), [np.inf]])
    return {"freq_{:d}".format(ii):
            np.logical_and(freqs > fsplit[ii], freqs <= fsplit[ii+1])
            for ii in range(len(fsplit)-1)}
```

Is it necessary to have the single band selections?  Those seem more like `enterprise_extensions` things rather than core `enterprise`.  Any thoughts @stevertaylor?

You also should add a test (`test_band_noise`?) that uses the band selection similar to the backend test here: https://github.com/nanograv/enterprise/blob/master/tests/test_gp_signals.py#L282.

vallis:
Currently seems to be about ephemeris? Closing this since code evolved too much—our bad, sorry.

PR #205, opened 2019-07-30
Determine parameters recursively in PTA.param

vallis:
This is needed, for instance, when a UserParameter depends on a hyperparameter.

```python
def scaledlogprior(x, scale=1):
    return -0.5*np.dot(x, np.dot(phi, x))*scale**2 - 0.5*np.log(np.linalg.det(phi)) # forget the 2pi

cphi = scipy.linalg.cholesky(phi).T
def scaledsampler(scale=1, size=4):
    return scale * np.dot(cphi,np.random.randn(size))

slp = parameter.Function(scaledlogprior, scale=parameter.Uniform(0.5,1.5))

cf = parameter.UserParameter(logprior=slp, sampler=scaledsampler, size=4)('eph_coefficients')
```

### DIFF
--- a/enterprise/signals/parameter.py
+++ b/enterprise/signals/parameter.py
@@ -96,7 +96,10 @@ class Parameter(object):
                 raise ValueError(
                     "You shouldn't give me my value when you're sampling me.!")
 
-            return self.prior(func=self._sampler, size=self._size, **kwargs)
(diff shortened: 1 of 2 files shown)


paulthebaker:
I'm excited to try this out!

PR #211, opened 2019-10-10
Reorganizing gp bases and priors.

Hazboun6:
This adds a few of the functions from [`enterprise_extensions`](https://github.com/nanograv/enterprise_extensions/) into the base package. It also reorganizes the Gaussian process functions out of `utils.py` and into their own sub modules. Backwards compatibility is retained by calling the functions explicitly in the namespace of `utils`, as suggested by @vallis.

### DIFF
--- a/enterprise/signals/gp_bases.py
+++ b/enterprise/signals/gp_bases.py
@@ -1,0 +1,230 @@
+# gp_bases.py
+"""Utilities module containing various useful
+functions for use in other modules.
+"""
+
+from __future__ import (absolute_import, division,
+                        print_function, unicode_literals)
+
+import numpy as np
+from enterprise.signals.parameter import function
+######################################
+# Fourier-basis signal functions #####
+######################################
+
+
+@function
+def createfourierdesignmatrix_red(toas, nmodes=30, Tspan=None,
+                                  logf=False, fmin=None, fmax=None,
+                                  pshift=False, modes=None):
+    """
+    Construct fourier design matrix from eq 11 of Lentati et al, 2013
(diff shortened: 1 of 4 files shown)


Hazboun6:
So in order for the `lint` test not to fail I had to `import *` within `utils.py` which I don't like to do. I tried importing the various function, but it didn't like that I had imported unused functions, and I couldn't fit the ignore error commands in there without going over the line limit :-) I'll work up some more tests soon...

vallis:
Thanks Jeff. Looks good, as soon as you can get the coverage up we'll merge. I don't mind the import star, although it means that all the bases, and not just the legacy ones, would be available from utils.

But re the import statement vs linting, could you do

```python
from module import (function, another_function,
                    another_function)
```

or even

```python
from module import function, another_function, \
                   another_function
```

and give the linting direction on the second line? Maybe that's what you tried already.

jellis18:
> So in order for the lint test not to fail I had to import * within utils.py which I don't like to do. I tried importing the various function, but it didn't like that I had imported unused functions, and I couldn't fit the ignore error commands in there without going over the lint limit :-) I'll work up some more tests soon...

You should be able to do something like
```python
from module import (function, another_function, # noqa: F401
                    another_function)
```
where `F401` is just a place holder for whatever flake8 error you are getting. This will ignore those kinds of errors in that import statement

Hazboun6:
Thanks @jellis18 . I actually has to do 
```python
from module import (# noqa: F401
                                         function, 
                                         another_function, 
                                         another_function)
```
So that works.

PR #200, opened 2019-05-31
Enterprise v3

paulthebaker:
This PR will take us to enterprise v3.0.0, officially dropping support for python 2.

PR #212, opened 2019-10-12
Add Black and reformat

jellis18:
This PR is quite large in that it touches probably every file in the code base. The main contributions are as follows

1. Added `make init` which will setup a virtual environment and install all of the dependencies in one step.
2. Add [black](https://black.readthedocs.io/en/stable/index.html) formatting. This has formatted all of the code.
3. Added a `make format` option which will run black on the repo. So this will do formatting automatically. 
4. Modified the `make lint` option that will check for black or flake8 errors but does not correct them.
5. Added flake8 and black commit hooks so it will automatically check for black or flake8 errors before allowing commits. This way all commits will be compliant.

The main reason for doing all of this is so that the code has one standard format and that format is automatic (i.e. `make format`). This will help out with CI as you don't have to wait for all of tests to run just to find out that you have a linting error.

This PR is only python3 compatible. It can probably be modified to cover python2 as well but I figured you all will be dropping python2 support soon. Also this PR sets the max line length to 120 instead of 80 which I think is more readable but this option can be set in the config files.


paulthebaker:
I would vote for reducing the max line length.  For instance this 118 character line seems too long:
```
        ans = (1.0 / np.sqrt(2)) * (ss.sph_harm(mm, ll, phi, theta) + ((-1) ** mm) * ss.sph_harm(-mm, ll, phi, theta))
```

There are a few simultaneous reformatting projects going on.  #200 which is a bit stale, but was meant to drop python 2.7 support.  #211 which moves some `enterprise_extensions` stuff to `enterprise` base as part of a larger `enterprise_extensions` reformatting.

We should combine these reformatting projects.  I think @Hazboun6's work is the most "disruptive", so maybe that should be the one PR to start from.

What do we think of at @jellis18 and @paulthebaker making PR's of their changes in to that branch?

Otherwise we could coordinate merges and or rebases to get all of this into master in separate PRs is some planned order.

jellis18:
@paulthebaker, I've merged my changes with @Hazboun6's [here](https://github.com/Hazboun6/enterprise/compare/jsh-add-new-gp-files...jellis18:formatting-jeff).

In terms of the line length. That is totally up to you guys.

Hazboun6:
This all sounds good to me. I am guessing I should have merged before my last set of changes though? My new tests are running, then I can work on merging with @jellis18 's branch.

jellis18:
@Hazboun6, I submitted a PR to your branch

PR #215, opened 2019-12-17
adding new custom backend selection function

siyuan-chen:
Hi all,
just adding a new convenience function for people to specify a list of backends that they want to use with `selections.Selection`. In principal it can recover the `by_backend` selection with argument None, and the `nanograv_backends` with argument list of the NG backends.

This is to help testing the use of ECORR on newer EPTA/PPTA data that could have some subband TOAs from the same observation, similar to the use in NANOGrav. And other selections people may want to use.

Tutorial:
selection = selections.Selection(selections.custom_backends(None))
ngb = ["ASP", "GASP", "GUPPI", "PUPPI"]
selection_ng = selections.Selection(selections.custom_backends(ngb))
cb = "ASP", "GASP", "GUPPI", "PUPPI", "NUPPI", "ROACH"
selection3 = selections.Selection(selections.custom_backends(cb))

Result:
selection should be the same as selection.by_backend
selection_ng should be the same as nanograv_backends
selection3 is a new custom selection


Hazboun6:
The changes that @siyuan-chen has made here seem reasonable. The IPTA analyses using some legacy backends need the pre-2000 ephemeris files, i.e. the setIII 1980 file. 

The coded selection addition is definitely an improvement over selections that are hard coded for specific backends. 

I'm not sure the flat power law is needed. This might be more appropriate in extensions, but do others have thoughts?

enterprise/signals/selections.py
```diff
@@ -119,6 +119,25 @@ def nanograv_backends(backend_flags):
...
+        """Selection function to split by custom backend flags only.
+        cb : list of str of the backends
+        use None to recover by_backend
+        use ["ASP", "GASP", "GUPPI", "PUPPI"] to recover nanograv_backends
+        """
+        nonlocal cb
+        flagvals = np.unique(backend_flags)
+        if cb is not None:
+            cb = list(np.atleast_1d(cb))
+            flagvals = filter(lambda x: any(map(lambda y: y in x, cb)), flagvals)
```
paulthebaker, line 133: Is this list comprehension more readable?
```suggestion
            flagvals = [flag for flag in flagvals if any(this_cb in flag for this_cb in cb]
```

paulthebaker:
I find list comprehension more readable than filter/map.  I made a suggestion in the diffs.

paulthebaker:
> I'm not sure the flat power law is needed. This might be more appropriate in extensions, but do others have thoughts?

I agree that the flat power law is better suited to `e_e`.

paulthebaker:
The ephemeris files are good.  If you make a PR with just the SSE files, I can merge.  Otherwise I can try to manually `git cherry-pick` the SSE commits, outside of the github workflow.

enterprise/signals/selections.py
siyuan-chen: I copied that line from `nanograv_backend`, so whatever looks nicer to people and passes the formatting checks, is good to me.

siyuan-chen:
> > I'm not sure the flat power law is needed. This might be more appropriate in extensions, but do others have thoughts?
> 
> I agree that the flat power law is better suited to `e_e`, but `broken_powerlaw`, `turnover`, etc. are all in `enterprise` proper. By that precedent maybe it does belong here.
> 
> If we want `enterprise` to only contain core functionality, we should move all of them to `e_e`.

I am happy to have those all in `e_e`; I was looking for where those other function were, to put the flat powerlaw in the same place.

@paulthebaker How does one do a PR with only selected elements? Originally, this PR was only supposed to be for the backend selection. But somehow newer commits were added to this PR. So, it didn't make sense for me to open another PR.

siyuan-chen:
Hi all, also closing this PR due to too many changes in the proposed merger. Will resubmit cleaner PRs mergers.

PR #217, opened 2020-01-14
Suggested addtions to enterprise for WidebandTimingModel

pennucci:
Additional dev/optimization could be done, specifically the ability to pass WidebandTimingModel the same sorts of options as is available for the normal TimingModel, and to speed up get_phi, but this is what I'm currently using.


vallis:
Thanks Tim. Can you update the files in your repo so they pass the linting/black check? I'll e-mail you what I get with black.

pennucci:
I updated the files; perhaps the merge should be tried again?

vallis:
See #219, which should fix the linting, and adds extra likelihood terms.

PR #218, opened 2020-02-20
Relative import issue

bencebecsy:
This resolves the issues I had with the relative errors previously in place. I'm not sure if this is system/python version specific or if there's a way to make relative imports work. But just changing them to `from ... import ...` style imports resolves the issue for me and I don't see a reason why that would break anything.


Hazboun6:
Looks like one tiny issue here @bencebecsy . The `flake8` test fails, which is really just for formatting:
`flake8 .
./enterprise/signals/utils.py:21:1: F401 'enterprise.signals as sigs' imported but unused
Makefile:42: recipe for target 'lint' failed
make: *** [lint] Error 1
The command "make lint" exited with 2.`

bencebecsy:
@Hazboun6 , I modified that import so it is ignored in the test. I've just realized that this is how it was done before. I'm not sure why, but I figured it's better than removing that import altogether. Let me know if you would prefer to remove it instead and I can do that.

PR #221, opened 2020-02-27
Adding sunssb position

Hazboun6:
Pulsar object now pulls in the newly available sunssb position vector for use in the SW model.


vallis:
Be sure to use the same version of black as the CI... there are minor infuriating differences between them.

Hazboun6:
By accident, it appears that I used the same version.

PR #224, opened 2020-04-14
Phase Shift Seed

Hazboun6:
Adding the ability to use a seed for the random phase shifts need for null hypothesis testing. The seed is added to an integer derived using the first TOA of the pulsar, so that each pulsar has a separate phase shift.


Hazboun6:
So the tests won't pass, but they pass locally on my machine. I've pasted the two kinds of error messages below. The only reference I can find to this is for an issue with `cython`, fixed in later version. 

@vallis, @stevertaylor and @paulthebaker Is there a specific reason that all of the dependencies have strict `==`, or just for safety's sake? I'm not sure which one could be the issue here, maybe it is `cython`. I might start adding`>=` to requirements one at a time, if that seems reasonable. 
```
======================================================================
ERROR: test_ecorr_sparse_ipta (tests.test_white_signals.TestWhiteSignalsPint)
Test of sparse ecorr signal and solve methods.
----------------------------------------------------------------------
ValueError: Cannot set the NumPy array 'base' dependency more than once
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/home/travis/build/nanograv/enterprise/tests/test_white_signals.py", line 413, in test_ecorr_sparse_ipta
    self._ecorr_test_ipta(method="sparse")
  File "/home/travis/build/nanograv/enterprise/tests/test_white_signals.py", line 378, in _ecorr_test_ipta
    assert np.allclose(N.solve(self.ipsr.residuals, logdet=True)[1], wd.logdet(), rtol=1e-8), msg
  File "/home/travis/build/nanograv/enterprise/enterprise/signals/signal_base.py", line 951, in solve
    mult = cf(other)
SystemError: <sksparse.cholmod.Factor object at 0x7f29bcf95780> returned a result with an error set
======================================================================
FAIL: test_like_corr (tests.test_likelihood.TestLikelihood)
Test likelihood with spatial correlations.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/nanograv/enterprise/tests/test_likelihood.py", line 266, in test_like_corr
    self.compute_like(npsrs=2, inc_corr=True)
  File "/home/travis/build/nanograv/enterprise/tests/test_likelihood.py", line 257, in compute_like
    assert np.allclose(eloglike, loglike), msg
AssertionError: Incorrect like for npsr=2, phiinv=partition
```

Hazboun6:
I changed the `==` in both `Cython` and `scikit-sparse` to `>=` and the checks passed. I'm now reverting back to using `numpy.random` instead of the Python module `random`. If checks pass I'll merge.

PR #232, opened 2020-10-01
Adding some flexibility to PhysicalEphemerisSignal

bvgoncharov:
Hi All, @danielreardon added inner Solar System planets to BayesEphem, which looks useful. I pulled Daniel's updates and I added some minor improvements to `enterprise`, so that there is flexibility on what BayesEphem components are included in a model. I also think it's better to replace kwargs of `PhysicalEphemerisSignal `, so that it takes priors rather than True/False. This way, people can choose priors on planet mass and orbit perturbations themselves, instead of using hardcoded values. 

Update 2020/10/01: I also added a possibility to fit for orbital elements of all planets, not just Jupiter and Saturn. Orbit partials were obtained with [this code](https://github.com/nanograv/bayes_ephem_project)


paulthebaker:
I don't understand the rational behind removing the default BayesEphem `PhysicalEphemerisSignal` kwargs.  A user can always pass their own kwargs when instantiating the signal to override the defaults.  Removing the defaults, makes it harder for users to implement BayesEphem effectively.  It also model breaks our unit tests which rely on the defaults being there.

bvgoncharov:
@paulthebaker , thanks for your reply. The only change for the user is to pass `d_saturn_mass=parameter.Uniform(...)` instead of `d_saturn_mass=True`. The same is true for `sat_orb_elements`, for all planets. If a person doesn't need a planet to be included, he doesn't need to pass anything, all defaults are zero. My take is that people who work on planetary ephemerides might want to try different priors now or in the future. Currently, priors are hardcoded, and a few years ago you told me yourself that hardcoding numbers is not the best way :) I also noticed the test failure, and I am happy to try to fix the test too. Otherwise, I am still OK with changing the kwargs back to True/False. Please let me know what you think.

The rest of the update just generalizes enterprise to all planets and their orbital elements.

paulthebaker:
As a warning, a big PR that makes many unrelated changes is unlikely to be merged.  In this case I think adding more functionality to BayesEphem is acceptable, but a lot of the selection stuff isn't really right for `enterprise` base.  Enterprise provides the `Selection` class and an example usage, but the many special selection types really should be implemented in a specific project.

The github interface doesn't let you `git cherry-pick` individual commits from a PR, so its all or nothing.

If you want to work on the BayesEphem stuff, I'd recommend starting a clean PR from a fork of the current `nanograv:master` branch and only include the BayesEphem stuff in that PR.  I suspect @vallis will have some strong opinions about how best to implement any changes to BayesEphem.  It might be more productive to discuss possible enhancements to BayesEphem in an Issue.

You are right that it't not ideal to have numerical values hardcoded in that way.  Implementing those defaults in a more readable way would be a welcome enhancement.  The main thing is that the default behavior remains the same.  Many existing analyses, including most "blocks" in enterprise extensions, instantiate BayesEphem with the defaults.  If the default behavior, changes it would break a lot of code.  Changing the tests to work with new defaults isn't the solution.

bvgoncharov:
@paulthebaker , I completely agree with you about selections. In fact, these selection updates are only displayed in commit history - there are no additional selections added in the current pull request. I really pulled the most recent `nanograv:master` and only changed `deterministic_signals.py` and `utils.py` to add these specific BayesEphem updates. There should be nothing else, although please correct me if I am wrong. 

As for keeping the default BayesEphem behavior - that makes sense, I will fix that and let you know.

bvgoncharov:
In the tab "Files changed", there are only the BayesEphem changes that I described. Thus, I assume, the rest is consistent with `nanograv:master`

bvgoncharov:
No problem, @vallis ! For the PPTA DR2 GWB paper, I actually implemented a code that allows doing model selection for each orbital element separately and I included all Solar System planets. For that, I introduced a "UniformMaskPrior" that allows to switch on and off individual orbital elements. The code is here: https://github.com/bvgoncharov/correlated_noise_pta_2020/blob/1990d3aa4f2d436f84fcfb19317b00b11590c230/ppta_dr2_models.py#L150
If that is interesting, I could help to move that to the main branch.

PR #236, opened 2020-10-15
removed global variables

AaronDJohnson:
4 variables outside functions removed from `enterprise.signals.utils`. `Pulsar` class in `enterprise.pulsar` should now load without `libstempo` installed.

### DIFF
--- a/enterprise/constants.py
+++ b/enterprise/constants.py
@@ -43,3 +43,9 @@ erg = sc.erg
 
 # other things
 DM_K = 2.41e-16  # for DM variation design matrix
+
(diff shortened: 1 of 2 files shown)


paulthebaker:
The `black` linter is failing, you can run `black --diff` to see how it thinks things should be formatted.

It looks like `e_ecl` is the relative angle between the Earth's ecliptic and the galactic equator, and `M_ecl` is the unit vector pointing in that direction.  Those seem more widely useful and could be moved to `enterprise.constants`

Can you add a comment to the t_offset definition labeling it as `# MJD 2010/01/01`.  I haven't checked, but the docstring says this.

PR #240, opened 2021-01-08
Updated call to Pulsar in case of timing_package='pint'...

pennucci:
when using paths to par/tim files to use pint.models.get_model_and_toas.

Using get_model_and_toas is 'safer' in that it automatically applies commands from the timing model (e.g., clock, SS ephem) as necessary.  This can still be overridden by the kwargs to Pulsar, but the default is to get them from the timing model.

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -26,9 +26,8 @@ except ImportError:
 
 try:
     import pint
-    import pint.toa as toa
-    import pint.models.model_builder as mb
-    from pint.models import TimingModel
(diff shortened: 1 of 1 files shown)


pennucci:
@Hazboun6 Can you review/merge this?

Hazboun6:
@pennucci This looks great. Thanks for bringing this to our attention and doing this so quickly!

Hazboun6:
@pennucci Sorry. I meant to merge ... Once the tests pass I'll merge.

Issue #238, opened 2020-11-22
PintPulsar needs to pull SSB vectors from Pint


Hazboun6:
Currently `PintPulsar` populates `_get_planetssb` and `_get_sunssb` with arrays of zeros. These are the vectors needed from the timing package in order to use `BayesEphem` and the solar wind models, respectively. These are available in `Pint`, but the code in `enterprise` needs to be changed in order to be able to do ephemeris and solar wind modeling using `Pint` as the basal timing package.

https://github.com/nanograv/enterprise/blob/master/enterprise/pulsar.py#L341

Hazboun6:
@vallis While #244 added the needed vectors, @AaronDJohnson's testing has not been successful. 

@AaronDJohnson can you resubmit a PR, and link it here, when you get a chance to investigate the issue up close?

Hazboun6:
Okay, we've diagnosed one major issue, which is that the `PintPulsar` object seems to lack a meaningful `pos_t` [attribute](https://github.com/nanograv/enterprise/blob/master/enterprise/pulsar.py#L350-L351), which is needed for the `BayesEphem` model. 
```Python
# TODO: pos_t not currently implemented
self._pos_t = np.zeros((len(self._toas), 3))
```

We need to define the vector from Pint, but I am not quite sure how to do that. The `libstempo`/`Tempo2` version looks like [this](https://github.com/nanograv/enterprise/blob/master/enterprise/pulsar.py#L463-L465):
```Python
self._pos_t = t2pulsar.psrPos.copy()
if "ELONG" and "ELAT" in np.concatenate((t2pulsar.pars(which="fit"), t2pulsar.pars(which="set"))):
    self._pos_t = utils.ecl2eq_vec(self._pos_t)
```

@paulray @aarchiba @scottransom is there an easy way to access this? I didn't see anything in the table that seemed pertinent. But I might not be recognizing an obvious name.

scottransom:
@Hazboun6 What is the `pos_t` attribute?  Is that the topocentric position of the pulsar?

Hazboun6:
@scottransom Yes, sorry for not saying that explicitly!

scottransom:
I'm pretty sure that is the `ssb_obs_pos` in the TOA table.  For instance:

```
In [5]: t.table['ssb_obs_pos'][0]
Out[5]: array([ 4.08342950e+07, -1.34788757e+08, -5.84439627e+07])
```

Hazboun6:
[...]

PR #244, opened 2021-01-19
Adding PintPulsar SSB Vectors

Hazboun6:
This PR tries to fix Issue #238 and pulls out the needed vectors from the Pint timing objects in order to do solar system ephemeris modeling. The unit tests pass, but the vectors seem to be off on the order of milliseconds. Before merging I'd like to hear from the Pint team about if this sort of difference is expected. Below are two images that show the vectors and the difference (`libstempo`-`Pint`) for Earth and Jupiter. These are for PSR J1911+1347 with the NANOGrav 12.5 year data set.

image

image


scottransom:
Hmmmm...  That seems further off than I would expect.  And am I reading things correctly in that the errors only seem to be in Y and Z and not X?  This will likely take some investigation.

Hazboun6:
@scottransom yeah, that's right regarding the coordinate differences. Below is just the x direction difference, which is more like ~50 nanoseconds. 
image

vallis:
Indeed, it would be good to understand this, although it won't affect BayesEphem, since the orbits of the planets are multiplied by M_planet/M_sun to yield mass-perturbation partials, and a few milliseconds out of thousands seconds will be a small correction.

Where does PINT get planetary ephemerides? Are they coordinate-transformed before being reported?

In other news, Jeff, could you merge master into the PR so you get the effects of the new continuous integration?

scottransom:
@vallis So the high-level PINT routine that is called to get the planet positions is compute_posvels() in toa.py.  That calls objPosVel_wrt_SSB() which is in solar_system_ephemerides.py.  However, the code that actually gets the positions and velocities from the JPL ephemerides is from the astropy coordinates get_body_barycentric() method.  So we probably need to dig in there (and perhaps compare with pyephem or some other codes).

Hazboun6:
@vallis I merged the recent CI changes. So you think it's okay to merge this PR and start testing `BayesEphem` runs with `Pint`?

Hazboun6:
@paulray, @luojing1211 These tests are failing on py3.8 for some weird error with `Pint`'s `urllib` download of ephemeris `bsp` files. Have either of you seen these with `Pint` before?

paulray:
Yeah, we get things like that sometimes. There is a good chance that re-running that test will fix it. There is an East Coast-wide internet issue today so that could be related.

aarchiba:
The code appears to be correct, basically inverting PINT's subtraction to get back the Astropy output:

https://github.com/nanograv/enterprise/pull/244/files#diff-27db36a2fdf65101108346718f2a100d56a8955c4cb7501dc9f4cafa0cf18384R366

versus

https://github.com/nanograv/PINT/blob/daea8fc7a5bc8b8de974907abb4d87d9bd47c613/src/pint/toa.py#L2032

We can (should) have a round-trip test with PINT to confirm that our generated vectors do agree with Astropy, but if you are depending on PINT already you could add a check of PINT and TEMPO2 SSB-planet vectors against Astropy to the test suite here.

Hazboun6:
Thanks @paulray rerunning did the trick!

Hazboun6:
@aarchiba thank you for the code cross check. That is very helpful.

PR #245, opened 2021-01-22
assorted minor changes

svigeland:
I made the following changes:

- Change to `get_lnprior()` function to fix deprecation warning
- Changed how pulsar distance is set based on the pulsar name to account for the fact that the 5-yr data set doesn't include the B or J prefixes in the pulsar names
- Updated the entry for PINT in the requirements.txt file to reflect the fact that PINT can be installed with `pip install pint-pulsar`


vallis:
Hmm... coverage still not reported correctly. We're assume it's an original sin (given this PR happened before the coverage process changes), and merge.

PR #246, opened 2021-01-22
CI with Github Actions

ark0015:


ark0015:
Because travis-CI is dropped, this won't pass those tests and probably the coveralls will not pass either. The new actions will automatically place the code on Codecov ([where PINT is](https://codecov.io/gh/nanograv)). The example coverage report is [here](https://codecov.io/gh/ark0015/enterprise/tree/4729e4df2e8e87e3914853b1a7470b5b83385676) and the example of the build passing with Github Actions is [here](https://github.com/ark0015/enterprise/actions/runs/504826294).

PR #247, opened 2021-01-25
coverage output in xml

ark0015:
This should output the coverage report in the correct format in addition to using the .coveragerc to exclude testing scripts, etc. from the coverage report.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -1,0 +1,27 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "75...100"
+  status:
+    default:
+      # basic
+        target: 50%
+        #threshold: 0%
+    project: on
+    patch: on
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,files,footer"
+  behavior: default
+  require_changes: no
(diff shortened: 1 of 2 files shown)


vallis:
So codecov is only reporting coverage, which is gathered by pytest. But where are we actually setting coverage requirements?

vallis:
Also look at the "out-of-date" branch warning...

ark0015:
> So codecov is only reporting coverage, which is gathered by pytest. But where are we actually setting coverage requirements?

I think that you can set the coverage to fail if it falls below some minimum (https://pytest-cov.readthedocs.io/en/latest/config.html with `--cov-fail-under=MIN`). Do we want a minimum amount of coverage? Was this what you were thinking?

> Also look at the "out-of-date" branch warning
I'm not sure where you are seeing this?

vallis:
Two things:
* codecov/project fails because coverage decreases... but that may be because previous numbers included the test code itself. So we may need to force-merge this PR.  
* The failure must be because of a default rule that coverage can't decrease... that may be too restrictive. Perhaps we could set a threshold at 85%, which would currently pass, then increase it gradually.

vallis:
Actually it looks like you already set a requirement at 50%?

ark0015:
> Actually it looks like you already set a requirement at 50%?

Yeah, I really thought I had it with that one... I'm not sure why. It might have to do with the default threshold? The target I think is the overall coverage and the threshold is the difference between the last one. I will try again with a larger threshold.

vallis:
Seems to be working now. Interesting that the codecov status was updated as soon the 3.7 CI-Test finished... since all three Pythons are running coverage, I suspect that the first one to finish is the one to report. Not a big problem.

PR #248, opened 2021-01-27
Attempting to turn off project and patch checks preventing merges

ark0015:
Since I'm not really sure why pull requests keep stalling with a `Expected -- Waiting for status to be reported` error on the Codecov steps, this should turn off the coverage checks, but still upload the coverage.
Example_Stalling

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -6,14 +6,8 @@ coverage:
   round: down
   range: "75...100"
   status:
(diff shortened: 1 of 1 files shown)


vallis:
I can turn off covariance checks in project settings, so we don't need to change the configuration file as here (which doesn't seem to be effective anyway...) We'll keep looking at coverage checks.

PR #249, opened 2021-02-05
Update deterministic_signals.py

AaronDJohnson:
Remove check that stops Bayes ephem from working with PINT. The code is in place to make it work now.


Hazboun6:
@AaronDJohnson it looks like there was a test of the exception that you removed. But we can just change that test to actually test the ephemeris model, which is what is done in the previous if statement.

PR #256, opened 2021-03-05
Add replicate() to enterprise/signals/utils.py

vallis:
The new function `utils.replicate(pta, ptac, p0)` works in tandem with `utils.get_coefficients()` to create replicated residuals conditioned on the data.

Here `pta` is standard marginalized-likelihood PTA, and `ptac` is a hierarchical-likelihood version of `pta` with `coefficients=True` for all GPs. The function:
* calls `utils.get_coefficients(pta, p0)` to get a realization of the GP coefficients conditioned on the data and on the hyperparameters in p0;
* calls `ptac.get_delay()` to compute the resulting realized GPs at the toas;
* adds measurement noise (including ECORR) consistent with the hyperparameters.
    
To use this (pending further development), you need to set `combine=False` on the `pta`/`ptac` GPs, and `method='sparse'` on the `ptac` `EcorrKernelNoise`.
    
Returns a list of replicated residuals, one list element per pulsar.

Somebody should also check my linear algebra...


jellis18:
This needs a test.

enterprise/signals/utils.py
```diff
@@ -27,11 +27,60 @@
     createfourierdesignmatrix_eph,
 )
 
-
 try:
     from sksparse.cholmod import cholesky
 except:
-    print("You'll need sksparse for get_coefficients() with common signals!")
+    print("You'll need sksparse for get_coefficients() and replicate()!")
```
jellis18, line 33: This will probably need to go away if we do a PR based on #258

pyproject.toml
```diff
@@ -18,6 +18,7 @@ exclude = '''
     | docs
     | .enterprise
     | enterprise/src
+    | enterprise/version.py
```
jellis18, line 21: I have no idea why but this should look like:
```python
exclude = '''
(
  /(
      \.eggs         # exclude a few common directories in the
    | \.git          # root of the project
    | \.hg
    | \.mypy_cache
    | \.tox
    | _build
    | buck-out
    | build
    | dist
    | docs
    | .enterprise
    | enterprise/src
  )/
  | enterprise/version.py
)
'''
```
This works. We probably also want to change the line in the CI that installs black to just do `python -m pip install -r requirements_dev.txt` since that uses a locked down version of black.

vallis: Indeed it works in my fork, thanks! I will close this PR, which has become a conflation of three different development efforts in my fork, and I will try to make separate PRs.

vallis:
Closing to split into modular PRs.

Issue #239, opened 2020-12-09
scikit-sparse install problems don't fail meaningfully


paulthebaker:
@siyuan-chen found an issue in a fresh `enterprise` install using `pip`.

`scikit-sparse` (sks) depends on the `SuiteSparse` C library.  If SuiteSparse isn't installed, `sks` can still install "successfully", but it will throw errors when used.

In [`LogLikelihood.__call__()`](https://github.com/nanograv/enterprise/blob/master/enterprise/signals/signal_base.py#L196) there's a `cholesky` call inside a `try` / `except` that catches **all** errors.  This means that if `sks` isn't linked to SuiteSparse correctly, the `sks` error will send the code to the `except` statement.  `LogLiklihood` will return `-inf` when it should crash and exit.  There's no indication to the user what went wrong.

The install problem can happen because [`requirements.txt`](https://github.com/nanograv/enterprise/blob/master/requirements.txt) installs `sks` with the `--no-binary` option.  This assumes the user has an existing install of SuiteSparse.  Installing the packaged binary of `sks` will use a generic BLAS, which is slower than a locally optimized version.  If you install packaged binaries of `numpy` and `scipy` via `pip`, they won't used optimized BLAS either.

There are few things to do:
 * additionally check `sks` linking when we [check that it is installed](https://github.com/nanograv/enterprise/blob/master/enterprise/signals/signal_base.py#L33) and give a meaningful error message
 * change the `requirements.txt` to use the packaged binary version of `sks`.  A `pip` install is already using generic BLAS for `numpy`... 🤷 
 * specify which errors that `try` / `except` is intended to catch, so the code exits on others. I think we want it to return `-inf` and continue if `Sigma` is singular

jellis18:
This is somewhat related to #257 and #258 and will somewhat be solved by [this PR](https://github.com/scikit-sparse/scikit-sparse/pull/64) I could possible add a check in here so the install will fail properly if the linked libraries aren't found.

PR #257, opened 2021-03-13
Updating github actions and requirements 

jellis18:
This PR does the following:

1. Revamps github actions tests to run on both linux and osx for python 3.6 - 3.8
2. Adds a build and release target to github actions. The build will build the wheels and release pushes them to PYPI. I don't know if there is a NANOGrav pypi account but that should probably be used. These build and release actions get triggered on a github release
3. Added `isort` pre-commit. This will ensure that imports are always consistent
4. Replaced `print` with loggers where appropriate
5. Removed all `__future__` since python 2.7 support is dropped.

There are still a few things that will need to be taken care of down the line. First off, we are still installing libstempo from github and not from a pypi release. I've been working with @vallis  on getting a pypi release ready so that will be along soon.

Next, scikit-sparse still requires `numpy` and `cython` to be installed beforehand because they don't have the right build dependencies set.  I made a PR to fix this but it may not get any attention (the repo hasn't been touched for a few years). Another option is to maybe fork it and use that for our `scikit-sparse` install (I already did this but it may be better to try to work through the scikit-sparse devs first).

Lastly, all of these changes are mainly meant for the `pip` path. Once there is a version on pypi we can then make a conda-forge version.

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 try:
     import libstempo as t2
 except ImportError:
-    logger.warning("libstempo not installed. Will use PINT instead.")
+    logger.warning("libstempo not installed. Will use PINT instead.")  # pragma: no cover
     t2 = None
 
 try:
@@ -33,7 +33,7 @@ try:
     from pint.residuals import Residuals as resids
     from pint.toa import TOAs
 except ImportError:
-    logger.warning("PINT not installed. Will use libstempo instead.")
+    logger.warning("PINT not installed. Will use libstempo instead.")  # pragma: no cover
     pint = None
 
 
@@ -574,4 +574,4 @@ def Pulsar(*args, **kwargs):
             os.chdir(cwd)
             return Tempo2Pulsar(t2pulsar, sort=sort, drop_t2pulsar=drop_t2pulsar, planets=planets)
     else:
-        logger.error("Unknown arguments {}".format(args))
+        logger.error("Unknown arguments {}".format(args))  # pragma: no cover
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -39,7 +39,7 @@ if pint is None and t2 is None:
     err_msg = "Must have either PINT or libstempo timing package installed"
     raise ImportError(err_msg)
 
-logging.basicConfig(format="%(levelname)s: %(name)s: %(message)s", level=logging.INFO)
+# logging.basicConfig(format="%(levelname)s: %(name)s: %(message)s", level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -118,7 +118,7 @@ class BasePulsar(object):
         """Sort data by time."""
         if self._sort:
             self._isort = np.argsort(self._toas, kind="mergesort")
-            self._iisort = np.zeros(len(self._isort), dtype=np.int)
+            self._iisort = np.zeros(len(self._isort), dtype=int)
             for ii, p in enumerate(self._isort):
                 self._iisort[p] = ii
         else:
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -6,8 +6,8 @@ import json
 import logging
 import os
 
-import astropy.units as u
 import astropy.constants as const
+import astropy.units as u
 import numpy as np
 from ephem import Ecliptic, Equatorial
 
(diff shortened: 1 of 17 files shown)


jellis18:
Does the coverage check now require that the coverage cannot drop at all? The coverage dropped 0.04% for this MR!

setup.py
```diff
@@ -10,14 +10,21 @@
...
 requirements = [
-    # TODO: put package requirements here
+    "numpy>=1.16.3",
+    "scipy>=1.2.0",
+    "ephem>=3.7.6.0",
+    "jplephem==2.6",
+    "healpy>=1.14.0",
+    "scikit-sparse>=0.4.2",
+    "pint-pulsar>=0.8.2",
+    "libstempo @ git+https://github.com/vallis/libstempo.git@master#egg=libstempo",
```
jellis18, line 20: This will change as soon as there is a release

vallis:
This looks great Justin. Thank you so much. Several things:

- The libstempo release (2.4.0) is now up, if you want to switch the install to that.
- We had problems with setting up the coverage requirements, so I had disabled it as a check. If you know how to fix it with reasonable criteria, please do.
- In addition to numpy/cython, the problem with scikit-sparse seems to be suitesparse. We need users to have a decent/fast version of that, otherwise there's no purpose in using scikit-sparse to begin with. Perhaps we can discuss this in #239.

jellis18:
@vallis if you just want coverage percent as a check I added that in this PR directly as part of the tests.

As for suitesparse I was unaware there were different builds of the main library itself. I thought it just had to do with whether or not a mkl numpy version was used to build scikit-sparse.

Unfortunately I seems nearly impossible to get mkl-based numpy though pip. I think the best approach may be to leave as is for the pip install path and just use conda if you care about speed. 

Doesn't look like conda-forge does anything special for [suitesparse](https://github.com/conda-forge/suitesparse-feedstock/tree/master/recipe)

How have you guys handled this in the past?

enterprise/pulsar.py
```diff
@@ -19,29 +19,28 @@
 except:
     import pickle
 
+logger = logging.getLogger(__name__)
+
 try:
     import libstempo as t2
 except ImportError:
-    print("Ooh, no libstempo?")
+    logger.warning("libstempo not installed. Will use PINT instead.")  # pragma: no cover
```
jellis18, line 27: @nanograv/enterprise These are not really needed if we have both `libstempo` and `PINT` as dependencies. Do we want to keep them both or have one as optional?

Hazboun6: I'm glad you changed them to `logger` warnings. I think we could make the case either way. People only need one of the packages to run enterprise. I think the only dependency on either package right now is within `enterprise.Pulsar`, so users can get a way with only using one. I guess that is a pretty rare situation for software. Maybe we just raise an `ImportError` if neither are present?

Hazboun6:
> @nanograv/enterprise I don't know how to turn off these checks above that aren't relevant...

I think we can delete them from `.github/.codecov.yaml`.  I assume we could just delete 
```
patch:
      default:
        target: auto
        threshold: 6%
```
Then there will be no patch coverage check. You are rerunning tests now, so I think those were the tests that weren't passing, is that correct?

jellis18:
@Hazboun6 the tests all pass now but yes we could remove that patch coverage check. What I don't know how to get rid of are these 
image

They don't run on pull requests and I don't even know where "lint" comes from as its not a target

Hazboun6:
> Actually I think I see now. Those were the old targets but aren't there anymore
Yeah, @jellis18  is this because GitHub Actions runs the old tests and the new tests in a PR?

Hazboun6 (approved):
@jellis18 This all looks great. Thanks for doing the hard work of getting the requirements set up correctly.

Hazboun6:
@jellis18 This is super weird. The Action [workflow](https://github.com/nanograv/enterprise/actions/runs/688035093) page doesn't even show those checks. Really annoying that we can't circumvent them somehow.

jellis18:
@Hazboun6, fixed it

enterprise/pulsar.py
jellis18: Well I think there are 3 options:
1. Don't require either as base dependencies and have a `pip install enterprise-pulsar[libstempo]` and `pip install enterprise-pulsar[pint]` path. I don't really like this because it could lead to broken installs because the package won't really work without one or the other.
2. Choose one as the default and add the other as an extra. Kind of like this one but not sure which should be default
3. Leave both as it is now. This is fine too

jellis18: Actually the whole `Pulsar` function/class is a pretty bad design. I don't think I knew much about class polymorphism at the time. Its too much to refactor it for this PR but it could definitely be better.

PR #259, opened 2021-03-24
Appending Pint's `toas` and `model` classes to `PintPulsar`

Hazboun6:
This PR adds the `pint.toas` and `pint.model` objects to `PintPulsar` objects in `enterprise`. These will be needed for nonlinear timing based on using `Pint`, but have been requested by @swiggumj as they are also needed for outlier analyses, which have not yet been carried out on the most recent NANOGrav dataset.

The functionality is already available for `Tempo2Pulsar` objects, which are dropped by default. My recollection is that is in part due to the fact that `libstempo` objects are not pickle-able since they are constructed using `Cython`. If developers feel like we should include an option to not include these then that can be added easily.


Hazboun6:
Okay, the dependencies won't install, but I just merged @master into this. Anyone know what went wrong? Is this similar to any issues in #257 ?

Hazboun6:
Okay, there seems to be some part of the `Pint` classes which won't pickle. Given that, I added a flag to drop these pieces. Really the flag gate keeps whether they get added, but I wanted to follow the `drop_t2pulsar` convention already established. So:
```
from enterprise.pulsar import Pulsar

psr = Pulsar('path/to/tim.tim','path/to/par.pa',timing_package='pint, drop_pintpsr=False)
```

will keep a copy of the `pint.toas` object as `psr.pint_toas` and a copy of the `pint.model` object as `psr.model`. These will be removed if pickling. Again this is following the convention already established that deletes `psr.t2pulsar` when pickling. 

I have added in a `logging.warning` that tells the user that these are being deleted when the pulsar is pickled using `psr.to_pickle`.

Hazboun6:
Okay, so this simple addition was needed for @swiggumj to work on the outlier analysis. I have definitely muddied the waters by needing to do some hot fixes of the installation, borrowing from @jellis18's PR. Given that PR #257 is doing a full reconnoitering of installations and CI, are people (@vallis , @paulthebaker ) ok with merging this PR for the time being so that the folks working on integrating the outlier analysis into `enterprise_extensions` and the timing analysis work flow can get started?

aarchiba:
The PINT TOAs object does pickle, but as a result of an astropy bug it needs a fix on unpickling (`toas.table = toas.table.group_by("obs")`). If this is an obstacle PINT's TOAs could acquire a custom pickler that just did that, I think. Models should pickle fine, as far as I can tell.

aarchiba:
The pickling issue for TOAs objects should go away when https://github.com/nanograv/PINT/pull/1021 is merged.

PR #261, opened 2021-03-27
Add .readthedocs.yaml

jellis18:
This should fix the docs build and allow for configuration via git instead of on RTD directly. I added a setting in RTD to build the docs on PRs as well so we will know if we break docs. 

You can look at the docs built from this PR in the checks.


jellis18:
Going to merge this in as I don't think there is anything to review

PR #268, opened 2021-05-21
add version info to summary

paulthebaker:
This adds enterprise version and Python version to the top of the `pta.summary()` output for better book-keeping.


paulthebaker:
I'm not sure if the import style is correct.  I guess the tests will tell me.

jellis18:
You need to run isort  or use the pre-commit hooks. They are amazing. `make init`

jellis18 (approved):

PR #282, opened 2021-06-16
Adding the fuzzy dark matter model.

brendrach:
- Added a FDM delay function to utils.
- Created a FDM delay test function.


paulthebaker:
This failed the linter.  You can run `black` locally to see what the problem is

vallis:
Code and test look good, thanks. I don't mind having this (which is rather self-contained) in enterprise core. Once the tests pass, this will be OK to merge.

vallis:
We'll make a new PR to remove BWM functionality from enterprise once it's settled in enterprise_extensions.

PR #289, opened 2021-08-30
Add partially shared Pulsar objects

vallis:
This PR adds Pulsar methods deflate(), inflate(), and destroy() that allow sharing Pulsar objects among processes on the same host/node.

The workflow is as follows:

- A prep-chef script loads all pulsar objects, deflates them, pickles them.
- All sous-chef scripts load the pickles, inflate the objects, use them.
- A kitchen-cleaner loads the pickles, destroys the objects, deletes the pickles.

This functionality requires Python 3.8 (so it is not tested in 3.6 and 3.7). It also requires dedicated testing on our production clusters.

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -5,6 +5,7 @@
 import json
 import logging
 import os
+import pickle
 
 import astropy.constants as const
 import astropy.units as u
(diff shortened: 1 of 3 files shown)


vallis:
Coverage fails on 3.6 and 3.7 because of the new shared-memory code. Should be able to exclude it along the lines of https://sobolevn.me/2020/02/conditional-coverage

AaronDJohnson:
I have tested this branch with both multiple serial tasks running on the same node and with MPI tasks running on the same node on a local cluster. Both versions are working as intended when using a Slurm scheduler.

PR #295, opened 2021-10-06
Adding Barycentric Pulsar Unit Vectors (pos_t) to PintPulsar

Hazboun6:
This PR adds in real values for the `Pulsar.pos_t` vector. This enables the use of this vector for SSB modeling. Changes include:

- Using the following call to Pint to get the position as a function of time, with proper motion corrections. 
```Python
if "AstrometryEquatorial" in model.components.keys():
    which_astrometry = "AstrometryEquatorial"
elif "AstrometryEcliptic" in model.components.keys():
    which_astrometry = "AstrometryEcliptic"

self._pos_t = model.components[which_astrometry].ssb_to_psb_xyz_ICRS(model.get_barycentric_toas(toas)).value
```
- Expanding the PintPulsar tests to include all of the tests that exist for `TempoPulsar`
- Removing an exception that prohibited `PintPulsar`s from being used in BayesEphem
- Expanding the deterministic signal tests to include a test of BayesEphem with `PintPulsar`

(sorry about the weird history here, but the actual code changes are pretty minimal)


Hazboun6:
@AaronDJohnson Tests passed locally for me, and assuming the pass here, it would be worth trying this branch out on the 12.5 year analysis to see if you can reproduce the BayesEphem results with `PintPulsar`s.

enterprise/pulsar.py
```diff
@@ -384,17 +388,22 @@ def _get_radec(self, model):
...
         vec = toas.table[obs_planet] + toas.table["ssb_obs_pos"]
         return (vec / const.c).to("s").value
 
     def _get_planetssb(self, toas, model):
         planetssb = None
         if self.planets:
             planetssb = np.zeros((len(self._toas), 9, 6))
-            # planetssb[:, 0, :] = self.t2pulsar.mercury_ssb
-            # planetssb[:, 1, :] = self.t2pulsar.venus_ssb
+            # planetssb[:, 0, :3] = self._get_ssb_lsec(toas, "obs_mercury_pos")
```
vallis, line 403: So we're leaving some of the orbits blank in planetssb?

Hazboun6: Some of the planet vectors are not available in Pint yet.  I put in a (hopefully) informative exception to that effect. I suppose I could clean up the code, but it seemed useful to have the others there. BayesEphem pulls by index, so we do need the zeros to space things out correctly. 

That said, the issue we had from the start was that `pos_t` was set to zero very quietly. Maybe I should add a logging message that mentions which planets are zero?

tests/test_pulsar.py
```diff
@@ -185,25 +186,41 @@ def setUpClass(cls):
...
+        assert len(self.psr.dmx) == 72, msg
+        assert self.psr.dmx["DMX_0001"]["DMX"] == np.float64("0.015161863"), msg
+        assert self.psr.dmx["DMX_0001"]["fit"], msg
 
-    def test_planetssb(self):
-        assert hasattr(self.psr, "planetssb")
 
-    def test_sunssb(self):
-        assert hasattr(self.psr, "sunssb")
+class TestPulsarPintRADEC(unittest.TestCase):
```
vallis, line 201: Same story as test_deterministic_signals.py... can we avoid duplicating tests for PINT?

Also, I think

```
def test_deflate_inflate(self):
         pass
```

needs to stay... unless deflation actually works with PINT?

enterprise/pulsar.py
vallis: How about we:

- add a comment to the BayesEphem code where planetssb is used, that planets X, Y, and Z are unavailable pending PINT enhancements...  
- make the empty planets NaN, not zero.

tests/test_deterministic_signals.py
```diff
@@ -177,16 +177,6 @@ def test_physical_ephem_model(self):
...
-            with self.assertRaises(NotImplementedError) as context:
-                eph1 = deterministic_signals.PhysicalEphemerisSignal(sat_orb_elements=True)
-                e1 = eph1(self.psr)
-
-                msg = "Physical Ephemeris model is not compatible with PINT "
-                msg += "at this time."
-                self.assertEqual(msg, str(context.exception))
-
 
 class TestDeterministicSignalsPint(TestDeterministicSignals):
```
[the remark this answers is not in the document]
Hazboun6: Yeah, oops. I missed that the subclassing took care of all of these tests. I did add a test for `setIII` here as well, since that is now our default. That did not include the saved vector check though.

tests/test_pulsar.py
Hazboun6: Alright, I think these are taken care of. There are a few other Pint specific tests. The DM variable check is different because the values come out as `numpy.float64`, not `longdouble`. The RA DEC import checks the `AstrometryEquatorial` option for `pos_t` and the no planet tag checks an exception in the `PintPulsar` class.

PR #297, opened 2021-10-12
Fix regression in 1d4975: Pulsar.backend_flags need to be properly sorted

vallis:
In commit 1d4975, I dropped a critical indexing step at line 266 in pulsar.py, which uses the toas sorting index to sort the backend_flags array. This resulted in scrambled selection masks when loading 12.5yr B1937 with PINT (but not with tempo2, which must do some sorting of its own). Aaron found this problem while testing PINT-BayesEphem. This one-line patch fixes the issue.

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -302,11 +302,14 @@ class PintPulsar(BasePulsar):
         self._sort = sort
         self.planets = planets
         self.name = model.PSR.value
+
         if not drop_pintpsr:
             self.model = model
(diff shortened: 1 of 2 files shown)


jellis18 (commented):
Since this seems pretty important, maybe you can add a test so this doesn't happen again by accident

vallis:
I'm also changing the content of PintPulsar.toas to match Tempo2Pulsar.toas (baricentered TDB toas). This may affect  physics at a very small level.

PR #306, opened 2021-11-03
Bump maximum Python to 3.10

vallis:


vallis:
Won't work for the moment because Scipy doesn't build with 3.10.

PR #315, opened 2022-01-27
Create combined EFAC + EQUAD noise object with tempo/tempo2 convention

vallis:
Instead of defining separate `MeasurementNoise` and `EquadNoise` signals, one would use `CombinedWhiteNoise(efac=..., log10_equad=..., selection=...)`.

### DIFF
--- a/enterprise/signals/signal_base.py
+++ b/enterprise/signals/signal_base.py
@@ -110,6 +110,9 @@ class Signal(object):
                 ret.append(p.name)
         return ret
 
+    def __repr__(self):
+        return "<Enterprise Signal object " + self.signal_id + "[" + ", ".join(p.name for p in self.params) + "]>"
+
     def get(self, parname, params={}):
         try:
             return params[self._params[parname].name]
--- a/enterprise/signals/white_signals.py
+++ b/enterprise/signals/white_signals.py
@@ -57,60 +57,55 @@ def efac_ndiag(toaerrs, efac=1.0):
     return efac ** 2 * toaerrs ** 2
 
 
-def MeasurementNoise(efac=parameter.Uniform(0.5, 1.5), selection=Selection(selections.no_selection), name=""):
-    """Class factory for EFAC type measurement noise."""
+@function
+def combined_ndiag(toaerrs, efac=1.0, log10_t2equad=-8):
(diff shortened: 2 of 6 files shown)


vallis:
To complete the transition and avoid confusion, EquadNoise(log10_equad=...) would be renamed TNEquadNoise(log10_tnequad=...). This will break all code that calls `EquadNoise`. Are we ready for that?

scottransom:
My opinion is to break things.  It seems that most people have been unknowingly using it incorrectly anyways!  Which is worse?

Hazboun6:
@vallis Would anything break if we changed the names of the Tempo-style parameter EQUAD to something else? Echoing something from a DWG telecon, maybe breaking the ability for a user to use old EQUAD values is the safest thing to do here. For instance changing the new EQUAD name so that the name comes out as `J1909-3744_log10_tequad` where the t stands for either Tempo or Timing (since all of the timing packages use that convention). Or another name more varied from T2EQUAD.

kdolum:
@Hazboun6, I'm not up on the current and proposed new set of parameter names, but I think what you're suggesting would address my issue.  My goal is that if someone has an old parameter file or notebook or set of instructions that refers to the old name, and they try to use it with new code, they get an error instead of a wrong answer.  Then they can hunt down the documentation of the change and update their files or process them with an old version or whatever.

vallis:
Yes, I think this is the right idea. I was already going to break things one way (using the “old style” equad objects with the wrong parameters), but we should do it both ways (also using “new style” objects with old chains).

This means the parameter name will be different from tempo/tempo2/pint, so an explicit mapping will be required. But that’s ok. I’m ambivalent about the name: I have tnequad for the temponest style; the tempo/tempo2/pint style could be t2equad (but pays undue homage to tempo2), tequad (but it’s hard to pronounce), trequad (for traditional), thequad (if you like puns)…

Opinions?

Hazboun6:
We could take a quick Slack poll in the timing channel?

vallis:
OK, so we're going with T2EQUAD after hearing the timers. Note that I've changed the interface again. EFAC and EQUAD (tempo-style) are now consolidated under `MeasurementNoise(efac=..., log10_t2equad=..., selection=...)`; the `signal_id` is measurement_noise. To use EFAC only, leave out `log10_t2equad`.

PR #318, opened 2022-05-23
Upgrading black to 22.3.0, editing code accordingly

vallis:
CI fails otherwise because black is incompatible with package click. The new black wants to make a few style changes, but they all seem reasonable or positive.


vallis:
Hmm... it seems the python-3.6 test failure is due to PINT. Will watch for a resolution there, or we could just drop 3.6.

vallis:
Dropping Python 3.6 to sidestep a PINT bug. 3.6 is EOLed and deprecated anyway.

Issue #321, opened 2022-08-31
inf/nan in deterministic signal waveform causes crash


paulthebaker:
## the problem
An `inf` or `nan` in the output of a deterministic signal causes a `ValueError` in `PTA.get_lnlikelihood` and crashes.  It should probably cause `PTA.get_lnlikelihood` to return `-np.inf` and continue.

The bad value in the waveform propagates into `TNr`.  Then the cholesky solve raises the `ValueError` during a check finiteness step.  We currently catch `LinalgError`s that occur during the cholesky factorization when `Sigma = TNT + phiinv` is bad.  In those cases we return `-np.inf` and continue.  This `ValueError` is not caught.

### when does it happen?
Waveforms can have bad values when parameters like `cos_theta` or `cos_inc` are out of bounds.  Then a calculation like `arccos(cos_theta)` returns a `nan`.

AM and SCAM proposals happily move out of bounds.  If the chain is near the edge, e.g. `cos_theta~1`, it's pretty easy to end up on the other side.  Additionally, these moves will occasionally propose "big" steps that are scaled by 10*sigma of the local covariance matrix.  Which can also go out.

### why had no one noticed until now?
In `PTMCMCSampler` the prior is always evaluated before the likelihood.  If the prior returns `-np.inf`, then the likelihood evaluation is skipped.  In almost all cases we set `cos_theta = Uniform(-1, 1)`, so the prior catches out of bounds values.  If we use a `Normal` prior, the prior won't forbid the out of bounds values.

## some solutions
1. Add `ValueError` to the `try`/`except` wrapper for the cholesky call.
1. Raise a new error in the waveform code, maybe `WaveformError`, and catch that when we call `get_TNr`
1. Run our own finiteness check on `TNr`.

`ValueError` is a pretty broad error, so (1) may catch things that really should crash.  Options (2) and (3) have the advantage that they don't waste cycles on the cholesky factorization, just for the solver to fail.  I think (2) is the most efficient, but my vote is for (3).  It seems simpler to set up.

vallis:
Thanks, interesting... but I think it's a good rule in numerical computing that nans should never be neutralized.

Otherwise (in this context) we may have situations where a bug is introduced in a waveform and it's never caught because sampling proceeds fine. Even the LinalgError -> -np.inf equivalence is a bit problematic, but what we're saying is that a covariance matrix that can't be inverted is probably in a region of parameter space that's not good for us.

[...]

PR #322, opened 2022-09-06
TruncNormal

paulthebaker:
This implements a truncated normal parameter as `enterprise.signals.parameter.TruncNormal`.  The truncated normal distribution matches the usual normal distribution up to normalization within a finite domain, and is zero outside of that domain.

This parameter class can be used to avoid the error described in #321.

`TruncNormal` handles scalar and vector parameters, including vectors where each element has a unique mean, deviation, and domain.  Unlike `Normal`, it cannot handle a full multivariate distribution defined by a covariance matrix.

This example code block shows the prior function and output of the sampler in agreement:
```python
N = 50_000
mu, sig, pmin, pmax = 0.50, 0.30, -1, 1
cosT = parameter.TruncNormal(mu, sig, pmin, pmax)("cosT")
samps = [cosT.sample() for ii in range(N)]

xs = np.linspace(-2, 2, 200)
pdf = cosT.get_pdf(xs)

plt.hist(samps, bins=30, histtype="step", density=True)
plt.plot(xs, pdf)
```
truncnorm

### DIFF
--- a/enterprise/signals/parameter.py
+++ b/enterprise/signals/parameter.py
@@ -307,12 +307,15 @@ def TruncNormal(mu=0, sigma=1, pmin=-2, pmax=2, size=None):
     """
 
     class TruncNormal(Parameter):
-        _norm = (
-            2
-            / np.sqrt(2 * np.pi)
-            / sigma
-            / (_erf((pmax - mu) / sigma / np.sqrt(2)) - _erf((pmin - mu) / sigma / np.sqrt(2)))
-        )
+        try:
+            _norm = (
+                2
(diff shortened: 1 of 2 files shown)


paulthebaker:
This should now correctly handle meta-parameters, by wrapping the initial normalization computation in a `try` / `except` statement.  If any of the four parameters are `Parameter` instances, this computation raises a `ValueError` and `_norm=None`.  The normalization must then be computed each time `get_pdf` is called.

There is also a new test that instantiates a `TruncNormal` with `mu` as a `Uniform` instance calls `get_pdf`

Issue #339, opened 2023-02-03
Segmentation fault in test suite


aarchiba:
I have a fresh install of the development version, and when I run the test suite I get a segmentation fault. The fault occurs at a different place in the test suite each time, once reaching 92% before dying, so it's possible the test suite might actually pass eventually without the problem being fixed.

I installed the development version following these instructions:
```
conda create -n ent_dev -y -c conda-forge python=3.9
conda activate ent_dev
conda install -y -c conda-forge enterprise-pulsar
conda remove enterprise-pulsar --force
conda install -y -c conda-forge black==22.3.0 flake8 sphinx_rtd_theme pytest-cov
pip install coverage_conditional_plugin
pip install -e .
```

Here is a partial list of the tests where the segfault has occurred:
```
tests/test_white_signals.py::TestWhiteSignalsPint::test_add_efac_tnequad Fatal Python error: make: *** [Makefile:69: test] Segmentation fault (core dumped)
tests/test_pta.py::TestPTASignals::test_parameterized_orf make: *** [Makefile:69: test] Segmentation fault (core dumped)
tests/test_pta.py::TestPTASignals::test_parameterized_orf Fatal Python error: make: *** [Makefile:69: test] Segmentation fault (core dumped)
tests/test_pulsar.py::TestPulsar::test_deflate_inflate Fatal Python error: Fatal Python error: Segmentation fault
tests/test_gp_signals.py::TestGPSignals::test_combine_signals Fatal Python error: Fatal Python error: Fatal Python error: Segmentation faultSegmentation faultSegmentation fault
```

paulthebaker:
Can you give us some system and OS info to see if we can get someone to reproduce it?

aarchiba:
I'm not quite sure what you would like, but here:

```
$ uname -a
Linux rhino 5.19.0-29-generic #30-Ubuntu SMP PREEMPT_DYNAMIC Wed Jan 4 12:14:09 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```
CPU:
```
model name	: Intel(R) Xeon(R) CPU E3-1220 v3 @ 3.10GHz
```

```
$ conda list
# packages in environment at /home/peridot/software/miniconda3/envs/ent_dev:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
[...]

Issue #332, opened 2023-01-31
Handle paths specified using Path objects


aarchiba:
Python has had for a while dedicated objects for working with paths (allowing things like `if parfile.exists():` or `timfile = parfile.with_suffix("tim")`). Currently at least the `Pulsar` factory cannot accept these:

```python
parfile = Path("file.par")
timfile = Path("file.tim")
Pulsar(parfile, timfile)
```
yields
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[9], line 1
----> 1 psr = Pulsar((parfile), (timfile))

File ~/software/miniconda3/envs/enterprise-development/lib/python3.10/site-packages/enterprise/pulsar.py:665, in Pulsar(*args, **kwargs)
    662         os.chdir(cwd)
    663         return Tempo2Pulsar(t2pulsar, sort=sort, drop_t2pulsar=drop_t2pulsar, planets=planets)
--> 665 raise ValueError("Unknown arguments {}".format(args))

ValueError: Unknown arguments (PosixPath('../enterprise_files/J0023+0923.nb.par'), PosixPath('../enterprise_files/J0023+0923.nb.tim'))
```

The user must apply `str` to each before passing them to `Pulsar()`. It should work fine with these objects.

aarchiba:
I think this could probably be fixed by broadening the check for strings in the `Pulsar()` function, but I'm hesitant to add any complication to the already complicated and unusual way it handles its arguments. Would it be okay to simplify the function? Specifically, accept exactly two positional arguments, with a preferred order but the ability to swap them if they are the wrong way around (optionally with a warning), plus some named keyword arguments, plus a `**kwargs` to be passed in to the eventual constructor?

I can put together a demo PR, though I can't run the test suite without segfaults (#339).

paulthebaker:
> I can't run the test suite without segfaults

Given what we've seen with the test segfaults, I would assume you could run `make lint` and `pytest tests/test_pulsar.py` successfully, even if `make test` fails.  After those run locally, we can count on the CI PR tests to catch anything else.

PR #340, opened 2023-02-10
Enterprise falls back to PINT if no TEMPO2

aarchiba:
This has three significant changes:

- If you run `Pulsar(par, tim)` without specifying PINT vs TEMPO2, one of two things will happen:
    - If you have TEMPO2 installed, you get a Tempo2 pulsar object as before.
    - If you don't have TEMPO2 installed, instead of an exception you now get a PintPulsar object.
- Almost all tests now run on PintPulsar objects if TEMPO2 is not available; the test suite passes and testing is reasonably complete.
    - Tests that require TEMPO2 (for example the par file uses `BINARY T2`) are skipped.
    - Tests that fail with PINT (inflate/deflate) are marked as expected to fail (`xfail`).
    - A few tests try to use both PINT and TEMPO2, skipping the TEMPO2 version if it's not available. (These tests were already run with both.)
- If PINT is not available, Enterprise continues to run using only TEMPO2 pulsar objects; the test suite likewise runs on almost all tests.

In short: you can use Enterprise transparently without TEMPO2. Also: a CI run now checks that Enterprise runs fine using PintPulsar for everything, and another checks that Enterprise runs fine without PINT. (Getting Enterprise to install when one or other dependency is missing is awkward but doable.)


aarchiba:
This is particularly valuable in light of #339 ; no seg faults if no TEMPO2.

aarchiba:
Still to do:

- [ ] Update CI so a run is done without TEMPO2 installed.

aarchiba:
Question: how do we make `libstempo` an optional dependency? The no-tempo run fails at `pip install -r requirements_dev.txt` because tempo2 isn't available (it didn't on my machine, for some reason). I believe I can fake this for now.

aarchiba:
The no-tempo2 installation is a bit of a hack, as we can't really tell pip "don't bother with this required dependency" and so it insists on installing `libstempo`, which fails if tempo2 is not available. 

Long-term, we could make the tempo2 dependency optional - `pip install enterprise[tempo2]` - but we can't really make that the default (except on conda?).

Short-term, we might be able to trick `libstempo` into claiming to install even though tempo2 isn't available. This could perhaps be done most easily with a `pip-install-fake-libstempo.py` that creates a fake pure-python `libstempo` of the correct version and does `pip install -e ` on that.

aarchiba:
To get adequate coverage of the patch I added a few tests; several of them are XFAIL because the behaviour of `Pulsar()` is bad but I hesitate to fix it here. #332 also calls for improvements to `Pulsar()`.

aarchiba:
We have three conditions for coverage:

1. The overall coverage of the entire set of CI runs must meet some threshold implemented by codecov.
2. The entire set of CI runs must cover some fraction of the lines in the PR, as implemented by codecov.
3. Every individual CI run must cover at least 85% (now 75%) of the code base; this includes CI runs where PINT or TEMPO2 are not available.

I think the third requirement is awkward and inconvenient from a CI point of view, although it is one that can be checked locally with `make test` (unlike the other two). I'm not sure what the right solution to this is.

Perhaps we could have two makefile targets: `make test` runs with a local coverage requirement, while `make test-ci` still gathers coverage information but does not fail if this single CI run has insufficient coverage. The existing codecov setup will make the PR fail CI if the overall coverage is insufficient.

aarchiba:
The changes to the code base are quite minor; just `pulsar.py`, mostly improved error handling. The bulk of the changes here are to the test suite, mostly just annotating tests with which program they need, but there are a few tests I had to convert to `pytest` style.

aarchiba:
I'm not sure what's up with the HTTP 404s - this occurs when astropy is trying to update its leap seconds file. It doesn't happen on my local machine, and it seems to only happen with the old python 3.7, pint 0.8.8, astropy 4.3.1 version; with astropy 5.2.1 the problem disappears. There were substantial upgrades to astropy's remote data fetching with astropy 5, so this may help explain why the problem only occurs with very old versions.

aarchiba:
Something has happened to prevent the coverage upload from the "no tempo2" version. I'm not sure what; maybe a passing network problem.

vhaasteren:
I do not like the hoops we are jumping through here to have libstempo installed by default, but to allow for a no tempo2 version. Our situation is perhaps somewhat unique that we have two independent packages, neither of which is ideal:

- Tempo2/libstempo are still a pain to install on some systems, despite progress in this regard
- PINT is properly packaged, but slow, and newer so less well-supported

Unfortunately, Python's packaging system doesn't directly support conditional dependencies based on the presence of other packages at install time. The closest built-in feature is the extra requires functionality, with those square brackets.

My suggestion to deal with this in a consistent way would be to
- Not have a required dependency on either libstempo or PINT
- Provide good documentation on the need for one of the two
- Perform a runtime check for either of the two packages

Of course the problem with this approach is that some build scripts and container scripts currently rely on libstempo being installed automagically, so if we make this change there will be things that break.

PR #345, opened 2023-05-04
Fixed name collision in class factory definition

### LINKED ISSUE #344: Class factories have the same name as the classes they return
Right now, class factories often have the same name as the class they return. In my PR for fast Kernel Ecorr (#343) I need a mechanism to check whether a class factory has been used before to show a warning message to the user only once. A decorator is a great way to do so. But the name collision used in Enterprise prevents that. Instead of jumping through hoops and do things the wrong way, I think it's better to use a better naming convention in Enterprise. Example that illustrates what goes wrong:

```
try:
    import fastshermanmorrison.fastshermanmorrison as fastshermanmorrison
except ImportError:
    fastshermanmorrison = None

import logging
logger = logging.getLogger(__name__)

class with_firstrun_counter:
    def __init__(self, func):
        self.func = func
        self.n_called = 0

    def __call__(self, *args, **kwargs):
        self.n_called += 1
        return self.func(*args, **kwargs)

    @property
    def first_run(self):
        return self.n_called == 1


@with_firstrun_counter
def EcorrKernelNoise():

    if EcorrKernelNoise.first_run and not fastshermanmorrison:
        logger.warning("Warning: install fastshermanmorrison for faster Kernel Ecorr")

    class EcorrKernelNoise(object):

        def __init__(self):
            pass

    return EcorrKernelNoise


if __name__=="__main__":

    EKN = EcorrKernelNoise()
    myobject = EKN()

    EKN = EcorrKernelNoise()
    myobject = EKN()

```

This causes an UnboundLocalError. By adding an underscore, like `class EcorrKernelNoise_(object) `, the code runs. Although valid Python in principle, the pattern used in Enterprise causes name collisions, and is best avoided.

vhaasteren:
This PR fixes #344. Inside the class factories, the class names now have a trailing underscore. This is not a user-facing change.

The main problem is that if you try to access attributes of the function (the class factory) itself, you get an `UnboundLocalError`, because the class that has the same name is in this scope and has not been defined yet. Making the names differ fixes that. A reason to access attributes of the class factory would be to get static variables, so one could for instance issue a warning only once, rather than every time you call a class factory. This can be done with global variables also, but it's less neat.

Currently, this PR is not exhaustive.

### DIFF
--- a/enterprise/signals/deterministic_signals.py
+++ b/enterprise/signals/deterministic_signals.py
@@ -13,13 +13,13 @@ from enterprise.signals.selections import Selection
 def Deterministic(waveform, selection=Selection(selections.no_selection), name=""):
     """Class factory for generic deterministic signals."""
 
-    class Deterministic(signal_base.Signal):
+    class Deterministic_(signal_base.Signal):
         signal_type = "deterministic"
         signal_name = name
         signal_id = name
 
         def __init__(self, psr):
-            super(Deterministic, self).__init__(psr)
+            super(Deterministic_, self).__init__(psr)
             self.name = self.psrname + "_" + self.signal_id
             self._do_selection(psr, waveform, selection)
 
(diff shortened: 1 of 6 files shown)


vhaasteren:
I'm closing this PR. It's not necessary

PR #348, opened 2023-05-17
Introduce utils.simulate()

vallis:
`simulate` will draw random residuals for all pulsars in `pta` by sampling all white-noise and GP objects for parameters `params`. 

* Requires GPs to have `combine=False`, and will run faster with GP ECORR.
* If `pta` includes a `TimingModel`, that should be created with a small `prior_variance`. Note that the variance applies to the entire design-matrix vector, and should be multiplied by `len(psr.toas)` if it is understood as the variance per individual residual. 
* This function can be used with `utils.set_residuals` to replace residuals in a `Pulsar` object.
* This function can also be used with `parameter.sample(pta.params)` to sample from the prior.
* Note that any PTA built from that `Pulsar` may nevertheless cache residuals internally, so it is safer to rebuild the PTA with the modified `Pulsar`.

### DIFF
--- a/enterprise/signals/signal_base.py
+++ b/enterprise/signals/signal_base.py
@@ -24,7 +24,6 @@ from sksparse.cholmod import cholesky, CholmodError
 from enterprise.signals.parameter import Function  # noqa: F401
 from enterprise.signals.parameter import function  # noqa: F401
 from enterprise.signals.parameter import ConstantParameter
-from enterprise.signals.utils import KernelMatrix
 
 from enterprise import __version__
 from sys import version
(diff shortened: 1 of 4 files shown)


vhaasteren:
I have a semi-related question because you are introducing `prior_variance` in `TimingModel`. I have also run into the need to use values for the prior_variance for the timing model parameters. Either for simulating data, but I also need it for analysis purposes. Problem is that the prior on timing model parameters is highly specific for each timing model parameter, and we have different parameters per pulsar. I ran into this challenge also when writing code to solve pulsars

What I don't like about using `prior_variance` in this particular way is that it is not flexible enough (I think, it's hard to see what BasisGP really can accept for priorFunction) to tailor the prior per timing model parameter per pulsar. I have an application in mind that requires that, so I have an interest in getting it right from the start.

In my previous work, my solution was to use the information in the par-files for this. The uncertainty column in the par-file is currently only used as output, even though libstempo and PINT do read it in. Actually, I'm not sure whether PINT can use it for anything else. But it's a perfect place in the par-file to write the prior width as input. I mean, the parameter values in the par-file are in reality also just the means of the prior in the analysis. People just haven't been thinking about it like that, but that's what it is.

So my suggestion is to use the uncertainty column in the par-file to specify the prior width.

vhaasteren:
Two more questions:

1) Regarding caching of residuals. When doing non-linear timing models, you can't cache the residuals like normal either. Those come from the timing package when you change parameters. How (and where?) is that implemented? Somehow we should be able to tell the caching descriptor that the residuals have changed and related quantities like TNr should be recalculated.

2) Will it still be faster with GP ECORR if you use my FastShermanMorrison code? I saw somewhere in here that it depends on the rNr solve of ShermanMorrison, which is super slow

enterprise/signals/signal_base.py
```diff
@@ -1212,6 +1211,89 @@ def solve(self, other, left_array=None, logdet=False):
...
+                return inv, np.sum(np.log(self))
+            else:
+                return inv
+        else:
+            try:
+                cf = sl.cho_factor(self)
+                inv = sl.cho_solve(cf, np.identity(cf[0].shape[0]))
+                if logdet:
+                    ld = 2.0 * np.sum(np.log(np.diag(cf[0])))
+            except np.linalg.LinAlgError:
```
vhaasteren, line 1286: Add a # pragma: no cover here?

enterprise/signals/utils.py
```diff
@@ -31,6 +32,73 @@
...
+            i = j
+
+        assert len(gp) == i
+    else:
+        for fmat, phi in zip(fmats, phis):
+            if phi is None:
+                gpresiduals.append(0)
+            elif phi.ndim == 1:
+                gpresiduals.append(np.dot(fmat, np.sqrt(phi) * np.random.randn(phi.shape[0])))
+            else:
```
vhaasteren, line 73: Add a # pragma: no cover here

enterprise/signals/utils.py
```diff
@@ -31,6 +32,73 @@
...
+                raise NotImplementedError
+
+    whiteresiduals = []
+    for delay, ndiag in zip(delays, ndiags):
+        if ndiag is None:
+            whiteresiduals.append(0)
+        elif isinstance(ndiag, ShermanMorrison):
```
vhaasteren, line 80: Not a good idea to check for an instance of ShermanMorrison. When 'fastshermanmorrison' is used, this will be a different type. Instead, perhaps duck typing can be used, like:

`if all(hasattr(ndiag, attr) for attr in ['_nvec', '_jvec', '_slices']):`

enterprise/signals/utils.py
```diff
@@ -31,6 +32,73 @@
...
+        pta.get_phi(params=params),
+    )
+
+    gpresiduals = []
+    if pta._commonsignals:
+        if sparse_cholesky:
+            cf = cholesky(sps.csc_matrix(phis))
+            gp = np.zeros(phis.shape[0])
+            gp[cf.P()] = np.dot(cf.L().toarray(), np.random.randn(phis.shape[0]))
+        else:
```
vhaasteren, line 57: Add a #pragma: no cover

vhaasteren (changes requested):
In general looks ok. I provided minor code feedback.

Question:

In the unit tests, right now the returned residuals are not actually checked, which makes the tests incomplete. Ideally we would generate an ensemble of realizations and compare that to the covariance matrix. That would be most realistic. It also sounds less feasible.

Also, my question regarding the timing model parameter priors still stand. IMO those should actually come from the timing package. I don't want to make that a show-stopper though.

Any thoughts?

PR #353, opened 2023-09-21
Percentage Point Functions for parameters

vhaasteren:
This PR adds the Percentage Point Function (PPF) to Enterprise parameters. The PPF is the inverse of the CDF, so it can be used to sample from the distribution. Some Nested Sampling implementations require the PPF as input: for instance MultiNest requires the "hypercube transformation". That is done with the PPF.

Now we can do something like:

```
from enterprise.signals.parameter import LinearExp

le = LinearExp(1, 3)("LEXP")
le.get_ppf(0.5)
```

The Truncated Normal PPF is not yet implemented, but `LinearExp`, `Normal`, and `Uniform` are there.


vhaasteren:
It seems that this requires some extra unit tests. Actually, the unit tests for the entire parameter class do not check for exceptions. I'll just update it to check for everything...

enterprise/signals/parameter.py
```diff
@@ -233,6 +261,17 @@ def NormalSampler(mu, sigma, size=None):
         return np.random.normal(mu, sigma, size=size)
 
 
+def NormalPPF(value, mu, sigma):
+    """Prior function for Normal parameters.
+    Handles scalar mu and sigma, compatible vector value/mu/sigma,
+    vector value/mu and compatible covariance matrix sigma."""
+
+    if np.ndim(sigma) == 2:
```
AaronDJohnson, line 269: Should this be >= 2? It looks like this ppf is for a 1D Gaussian, and the class factory for Normal can take more than 2 values for mu, sigma. Alternatively, we could implement a multivariate Gaussian ppf.

AaronDJohnson:
@vhaasteren Is there anything that needs to be done on this one?

vhaasteren:
No, is ready to go

enterprise/signals/parameter.py
vhaasteren: Ah! I totally missed this comment. Yes, good point, I made the change. PPF for multivariate Gaussian is indeed possible,  but let's postpone that for now

PR #352, opened 2023-09-21
Fixed the TypeError in test_vector_parameter_like

vhaasteren:
The integration tests fail for python 3.9 and 3.10, because the generator object is not converted to a list or tuple in the test function `test_vector_parameter_like`. See issue #354.

This PR fixes #354 by converting the generator object to a list first.

Issue #355, opened 2023-09-22
ECORR incorrectly only applied when `if slc.stop - slc.start > 1` in an epoch


vhaasteren:
In ShermanMorrison (and BlockMatrix?), the ECORR signal is only applied to epochs where there is more than 1 observation. Although the reasoning seems to make sense that this would otherwise just be an EQUAD at that epoch, this is not a correct way to restrict application. The ECORR signal in our generative model affects an observation whether or not there are more observations in that epoch. It _always_ affects the observation, whether or not it is distinguishable from an EQUAD or not.

So, for instance, in the below code we should change the `if` condition to `> 0`, so allow for single observations.

```
    def _get_logdet(self):
        """Returns log determinant of :math:`N+UJU^{T}` where :math:`U`
        is a quantization matrix.
        """
        logdet = np.einsum("i->", np.log(self._nvec))
        for slc, jv in zip(self._slices, self._jvec):
            if slc.stop - slc.start > 1:
                niblock = 1 / self._nvec[slc]
                beta = 1.0 / (np.einsum("i->", niblock) + 1.0 / jv)
                logdet += np.log(jv) - np.log(beta)
        return logdet

```

PR #356, opened 2023-09-22
Allow for index arrays in ShermanMorrison

vhaasteren:
This one solves #319 by allowing `EcorrKernelNoise` to work with index arrays, rather than slice objects. The function that is changed now is `quant2ind`, which now by default returns an list of index arrays.

The `fastshermanmorrison` package with Cython code is also compatible with this change, in case anyone is using that. Just upgrade it with pip

Thanks to Bjorn Larsen for providing a first outline of this modification.

**TODO:**
- [x] Modify quant2ind to return index arrays (and slices)
- [x] Fix failing tests in test_white_signals
- [x] Change `_setup_sparse` and `_get_ndiag_sparse` in `EcorrKernelNoise` to allow for index arrays
- [x] Update fastshermanmorrison-pulsar for full compliance
- [x] Added unit tests that shuffle TOAs for `EcorrKernelNoise`

Depends on #352 for codev report


enterprise/signals/signal_base.py
```diff
@@ -1188,12 +1190,12 @@ def _get_logdet(self):
...
-        for slc, block in zip(self._slices, self._blocks):
-            if slc.stop - slc.start > 1:
-                cf = sl.cho_factor(block + np.diag(self._nvec[slc]))
+        for idx, block in zip(self._idxs, self._blocks):
+            if len(idx) > 1:
+                cf = sl.cho_factor(block + np.diag(self._nvec[idx]))
                 logdet += np.sum(2 * np.log(np.diag(cf[0])))
             else:
-                logdet += np.sum(np.log(self._nvec[slc]))
+                logdet += np.sum(np.log(self._nvec[idx]))
```
AaronDJohnson, line 1198: This line is missing in the tests

enterprise/signals/utils.py
```diff
@@ -767,26 +767,37 @@ def create_quantization_matrix(toas, dt=1, nmin=2):
...
     return inds
 
 
+def indices_from_slice(slc):
+    """Given a slice object, return an index arrays"""
+
+    if isinstance(slc, np.ndarray):
+        return slc
+    else:
+        return np.arange(*slc.indices(slc.stop))
```
AaronDJohnson, line 798: missing tests

enterprise/signals/utils.py
```diff
@@ -767,26 +767,37 @@ def create_quantization_matrix(toas, dt=1, nmin=2):
...
     inds = []
     for cc, col in enumerate(U.T):
         epinds = np.flatnonzero(col)
-        if epinds[-1] - epinds[0] + 1 != len(epinds):
-            raise ValueError("ERROR: TOAs not sorted properly!")
-        inds.append(slice(epinds[0], epinds[-1] + 1))
+        if epinds[-1] - epinds[0] + 1 != len(epinds) or not as_slice:
+            inds.append(epinds)
+        else:
+            inds.append(slice(epinds[0], epinds[-1] + 1))
```
AaronDJohnson, line 788: missing tests

enterprise/signals/signal_base.py
vhaasteren: As you can see, the line it replaces was also missing in the test. This is just a replacement of slc with idx.

In order for us to cover this line, we need to add a NANOGrav test set that has only a single observation in an epoch.

My vote would be to leave the test coverage as it is, because IMHO this if statement should disappear. Even if len(idx)==1 we should include the ecorr. See issue #355

enterprise/signals/utils.py
vhaasteren: Thanks, you are right. I'll add it

enterprise/signals/utils.py
vhaasteren: Thanks, I'll add it

vhaasteren:
@AaronDJohnson, unit tests for the extra lines are done. As commented on the code, the remaining two lines that aren't covered have never been covered and IMHO shouldn't even exist. I vote to leave those like this

AaronDJohnson (approved):

PR #361, opened 2023-09-29
Add MockPulsar class, for easy use of Mock data with Enterprise

vhaasteren:
This adds a very rudimentary MockPulsar class: a class that can be easily instantiated without a timing package (PINT/Tempo2). All Enterprise models can be run on this type of pulsar object, but it can be instantiated from mock data very quickly. This is useful for anyone who wants to use Enterprise on fake data. Previously, one had to rely on fake par/tim files, created through libstempo.toasim or otherwise.

Note: this is coded up as part of Enterprise, and not Enterprise_extensions, because the pulsar class is essential for usage of Enterprise. Allowing Enterprise to be useful without a timing package was a big plus, but with libstempo/PINT optional I guess that's less of an advantage now

Todo:

- [x] Create unit tests
- [x] Add Astrometry parameters, aside from only Spindown

Additional todo, perhaps not part of this PR:
- Add helper functions to enterprise_extensions so this class can be instantiated easily
- Resolve unit disparity of PMRA / PMDEC between Tempo2 and PINT
- Add documentation for Tempo2Pulsar and PintPulsar

As in the test suite, a typical way to add this pulsar is:

```
toas = np.linspace(53000.0, 58000.0, 4005)
flags = {"f": np.array(["nosystem"] * 4005), "fe": np.array(["nofrontend"] * 4005)}
decj, raj = (0.16848694562363042, 4.9533700839400492)
# Can also define elong/elat

psr = MockPulsar(
    obs_times_mjd=toas,
    raj=raj,
    decj=decj,
    ssbfreqs=1440.0 * np.ones_like(toas),
    residuals=np.zeros_like(toas),
    toaerrs=1e-6 * np.ones_like(toas),
    sort=True,
    flags=flags,
    telescope="GBT",
    spindown_order=2,
    inc_astrometry=True,
)
```

This pulsar now has `psr.name == 'J1855+0939'`, and the design matrix contains astrometry and spin models (so 8 parameters). Residuals can be set with:

```
psr.set_residuals(residuals)
```

This pulsar is extremely fast to make, and can be combined with most Enterprise functionality. Astrometry parameters depend on Astropy (optional dependency).


vhaasteren:
This is still ready to be reviewed

PR #365, opened 2023-11-09
Modifications to the Pulsar class

vhaasteren:
This PR does not add any new functionality or change any functionality. Instead, it slightly refactors the Pulsar class in a way that makes it easier to create derived classes without code replications.

Modifications:
 - [x] the classes `PintPulsar` and `Tempo2Pulsar` classes now have attributes `drop_not_picklable`, which initially was implemented in the base class.
 - [x] The par and tim files are also saved in strings as part of the pulsar object.
 - [x] Designmatrix units are saved
 - [x] The requested timing package in `Pulsar` is attempted to be used inside of a try block
 - [x] Add extra unit tests for full coverage

These modifications allow, for instance, for the [`h5pulsar`](https://github.com/vhaasteren/h5pulsar) package to be used as a drop-in replacement instead of the base Enterprise pulsar classes. The `h5pulsar` derives it's pulsar classes from Enterprise without code duplication. The modifications in this PR are completely independent from that package though, and simply make the code more modular.


AaronDJohnson:
@vhaasteren does storing the string data change the memory requirement for the object significantly?

vhaasteren:
> @vhaasteren does storing the string data change the memory requirement for the object significantly?

Significantly is a relative term. The tim files for some pulsars are several MB in size. Must be around 20MB for the largest pulsars in the IPTA? I didn't even think about it, but I suppose for an entire PTA it could add up.

Currently, it's stored if we set `drop_t2pulsar=False` or `drop_pintpsr=False`. Are you perhaps suggesting to make that a different option?

AaronDJohnson (approved):

PR #370, opened 2023-11-30
Properly address units in PintPulsar position

### LINKED ISSUE #363: PINTPulsar does not check for coordinate units
I am trying to read in non-nanograv and non-PTA data, where the position of the pulsar is given in equatorial coordinates (as opposed to ecliptic coordinates used in nanograv), where RA is in units of hhmmss. 

Then enterprise pulsar object does not check for units, assuming they are supplied in radians, which results in an incorrect position used for these pulsars.

vhaasteren:
This fixes #363

By using astropy.units to convert to units.rad, it doesn't matter what units PINT returns, the position is read in radians.

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -399,12 +399,13 @@ class PintPulsar(BasePulsar):
 
     def _get_radec(self, model):
         if hasattr(model, "RAJ") and hasattr(model, "DECJ"):
-            return (model.RAJ.value, model.DECJ.value)
+            raj = model.RAJ.quantity.to(u.rad).value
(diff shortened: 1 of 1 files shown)


vhaasteren:
Also note that the designmatrix and model parameter units are not addressed by this PR. That is a separate issue that should be addressed in a more comprehensive PR

vhaasteren:
This one depends on #382 for the added test to pass. So @tcromartie, this last test I added was missing (uncovered lines). If it had been there from the start, your PR #382 would have never been happening.

Other than that, this PR is ready

AaronDJohnson (approved):

PR #373, opened 2023-12-23
Merge EPTA version part 1

siyuan-chen:
This pull request is part 1/2 to merge the EPTA enterprise into the NANOGrav version.

It adds more selection functions, for EPTA work and also a general generator function to use any flag with any value as selection criteria.
custom_backends_dict() takes a dictionary of {"flagname": "value"} and returns a selection based on the dictionary.

### DIFF
--- a/enterprise/signals/gp_bases.py
+++ b/enterprise/signals/gp_bases.py
@@ -219,7 +219,6 @@ def createfourierdesignmatrix_eph(
 
 @function
 def createfourierdesignmatrix_chromatic(toas, freqs, nmodes=30, Tspan=None, logf=False, fmin=None, fmax=None, idx=4):
-
     """
     Construct Scattering-variation fourier design matrix.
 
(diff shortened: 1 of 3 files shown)


vhaasteren:
Hi @siyuan-chen, this looks great. Aside from the fact that you need to lint the code first (use black), I have one comment:

I recently created a very similar selection function for my IPTA data combination project. Since I didn't want to touch the tim files, I needed a function that could work on multiple flags. So for ECORR, you can do:

```python
ecorr_sel = selections.Selection(create_selection_stag("ecorr", {('B', 'f'): None}))
```

('stag' is for 'staggered', so with a fallback. Not a great name, I know)

If you only want to use NANOGrav backends, it'd be:

```python
ecorr_sel = selections.Selection(create_selection_stag("ecorr", {('B', 'f'): ["ASP", "GASP", "GUPPI", "PUPPI", "YUPPI"]}))
```

or just

```python
ecorr_sel = selections.Selection(create_selection_stag("ecorr", {'f': ["ASP", "GASP", "GUPPI", "PUPPI", "YUPPI"]}))
```

The syntax of what I wrote seems to be a superset of what you wrote, so all EPTA code should work with it. Is it an idea to update it? My code is very ugly though (which is why I didn't put it out there yet), but I'm happy to share if you don't judge.

PR #290, opened 2021-08-30
Introduce experimental BasisCommonGP2 and FourierBasisCommonGP2 that accept selections

vallis:
For instance, to implement independent correlated GPs on each telescope, one could do
```
rn = gp_signals.FourierBasisCommonGP2(pl, orf, selection=Selection(selections.by_telescope), Tspan=Tspan)
```
Note that it is currently necessary to set `Tspan`.

The PR also adds the method `telescope()` to pulsar, and `by_telescope()` to selections.

PR #374, opened 2023-12-23
Merge EPTA version part 2

siyuan-chen:
This pull request is part 2/2 to merge the EPTA enterprise into the NANOGrav version.

It adds more createfourierdesignmatrix functions, for EPTA work and also a general function for chromatic signals. And also associated changes in other functions to pass more arguments.

### DIFF
--- a/enterprise/signals/gp_bases.py
+++ b/enterprise/signals/gp_bases.py
@@ -13,9 +13,12 @@ from enterprise.signals.parameter import function
 __all__ = [
     "createfourierdesignmatrix_red",
     "createfourierdesignmatrix_dm",
+    "createfourierdesignmatrix_dm_tn",
     "createfourierdesignmatrix_env",
     "createfourierdesignmatrix_ephem",
     "createfourierdesignmatrix_eph",
+    "createfourierdesignmatrix_chromatic",
+    "createfourierdesignmatrix_general",
 ]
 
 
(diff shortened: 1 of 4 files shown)


enterprise/signals/gp_bases.py
```diff
@@ -124,6 +127,42 @@ def createfourierdesignmatrix_dm(
...
+    :param toas: vector of time series in seconds
+    :param freqs: radio frequencies of observations [MHz]
+    :param nmodes: number of fourier coefficients to use
+    :param Tspan: option to some other Tspan
+    :param pshift: option to add random phase shift
+    :param fref: reference frequency [MHz]
+    :param logf: use log frequency spacing
+    :param fmin: lower sampling frequency
+    :param fmax: upper sampling frequency
+    :param modes: option to provide explicit list or array of
```
vhaasteren, line 148: The parameter `idx` needs a docstring

enterprise/signals/gp_bases.py
```diff
@@ -124,6 +127,42 @@ def createfourierdesignmatrix_dm(
...
+    """
+
+    # get base fourier design matrix and frequencies
+    F, Ffreqs = createfourierdesignmatrix_red(
+        toas, nmodes=nmodes, Tspan=Tspan, logf=logf, fmin=fmin, fmax=fmax, pshift=pshift, modes=modes
+    )
+
+    # compute the DM-variation vectors
+    Dm = (fref / freqs) ** idx * np.sqrt(12) * np.pi / 1400 / 1400 / 2.41e-4
+
```
vhaasteren, line 162: The matrix `Dm` is normalized differently from the one for `createfourierdesignmatrix_red`. This is not described in the docstring, and those factors are not explained anywhere.

Please make this consistent.

enterprise/signals/gp_bases.py
```diff
@@ -232,15 +273,82 @@ def createfourierdesignmatrix_chromatic(toas, freqs, nmodes=30, Tspan=None, logf
...
 
     # compute the DM-variation vectors
     Dm = (1400 / freqs) ** idx
 
     return F * Dm[:, None], Ffreqs
+
+
+@function
+def createfourierdesignmatrix_general(
+    toas,
```
vhaasteren, line 296: Question: is this not just a superset of the regular `createfourierdesignmatrix`? In other words, could it replace it?

vhaasteren (changes requested):
It needs to be linted and some docstrings need to be added. But overall this looks fine. Thanks!

enterprise/signals/gp_bases.py
achalumeau: Yes, it is a superset of "createfourierdesignmatrix_red", "..._dm", "..._dm_tn" and "..._chromatic". Where the "chromatic" is a superset of "dm" by the way.

The "general" also allows to mask toas from chosen flags. It is pretty convenient as it is the only current way to make a selection on spatially correlated common signals (as far as I know).

vhaasteren: Actually, selecting on common signals is in another PR as well. Have a look:

https://github.com/nanograv/enterprise/pull/290

achalumeau: Missed that, very nice ! It still need to be merged if I've understood well.

vhaasteren: Correct. I wanted to review it the other day, but I wasn't sure whether it was relevant since it had been 2 years. I should ping @vallis again regarding it.

enterprise/signals/gp_bases.py
siyuan-chen: added docstring for the parameter `idx`

enterprise/signals/gp_bases.py
siyuan-chen: It is the normalization difference between the enterprise and temponest default amplitude normalization for DM power law. There is the 12pi^2 from the definition of A, the scaling to 1MHz vs 1400 MHz and the DM constant. I added a bit of comments to this in the code.

vhaasteren: Understood, if it's a definition matter we should leave it. Looking up the origin, this is a really pointless normalization though. This 12pi^2 comes from the definition of the characteristic strain h_c in the red noise spectral density. It has absolutely nothing to do with DM, and it was introduced by mindlessly copying code from the definition of red noise to DM. It makes me cringe

vhaasteren:
Review completed @siyuan-chen, I'm fine with it all. If all the linting is done we can merge to dev.

vhaasteren (approved):

PR #376, opened 2024-02-14
Faster sparse matrix Cholesky

vhaasteren:
This PR changes the way the sparse matrix Cholesky is carried out. The first time the likelihood is called, a regular

```python
self.cf = cholesky(Sigma_sp)
```

is done. However, every subsequent call only the numerical decomposition is carried out. The analytical sparse decomposition takes a significant amount of time for our decompositions, so there is a speedup by just doing an 'update':

```python
self.cf.cholesky_inplace(Sigma_sp)
```

For the NANOGrav 15yr set, this alone reduces the model_3A time on my laptop from 370ms to 280ms. Additional speedups might be possible in the construction and decomposition of 'phi'

### DIFF
--- a/enterprise/signals/signal_base.py
+++ b/enterprise/signals/signal_base.py
@@ -223,7 +223,7 @@ class LogLikelihood(object):
                 try:
                     Sigma_sp = TNT + sps.csc_matrix(phiinv)
 
-                    if hasattr(self, 'cf_sp'):
+                    if hasattr(self, "cf_sp"):
                         # Have analytical decomposition already. Just do update
                         self.cf_sp.cholesky_inplace(Sigma_sp)
                     else:
--- a/enterprise/signals/signal_base.py
+++ b/enterprise/signals/signal_base.py
@@ -221,9 +221,17 @@ class LogLikelihood(object):
 
             if self.cholesky_sparse:
                 try:
-                    cf = cholesky(TNT + sps.csc_matrix(phiinv))  # cf(Sigma)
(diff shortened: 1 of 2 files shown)


vallis:
Looks ok. Perhaps it would be good to add a test that the results are invariant? Call the likelihood for two parameter sets, then erase the cached cholesky and do it again for the second parameter set.

Also, what's going on with the Mac OS tests?

AaronDJohnson:
It looks like `sksparse` faiiled on the Mac builds because it didn't install `suitesparse` for some reason.

AaronDJohnson:
It looks like the Mac builds failed to install `suitesparse` @vallis. I'm rerunning one of them to see if it's on their end.

vhaasteren:
One of the headers couldn't be found (cholmod.h), so perhaps they repackaged it? There were major modifications to the build system it says in the [changelog](https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev/ChangeLog).

Anyway, good idea regarding adding the extra tests. Once I figure out how to erase the cache I'll add those to this PR

AaronDJohnson:
Rerun these tests after merging the new dev changes to fix Mac CI :)

vhaasteren:
Great, they pass. I'll still add the tests though, so don't merge yet

vhaasteren:
Ok, I finally wrote the check @AaronDJohnson. This is ready for review/merge

AaronDJohnson (approved):
Looks good to me!

PR #379, opened 2024-03-04
Allow efac/log10_tneqad to be None in ndiag functions

mattpitkin:
In [white_signals.py](https://github.com/nanograv/enterprise/blob/master/enterprise/signals/white_signals.py), the `*_ndiag` functions fail if they get passed `None` values. This happens, for example, if the `white_noise_block` in
[enterprise_extensions](https://github.com/nanograv/enterprise_extensions/blob/master/enterprise_extensions/blocks.py) has `vary=False`, and therefore efac/equad/ecorr are set to `Constant` parameters, which have default values of None.

The two potential fixes are the one in this PR, i.e., allow `None` values, which for `efac_ndiag` is equivalent to having `efac=1.0`, and set `tneqad_ndiag` to zeros if passed `None. The other fix would be to specify values for the `Constants` in the `white_noise_block` in
[enterprise_extensions](https://github.com/nanograv/enterprise_extensions/blob/master/enterprise_extensions/blocks.py), e.g.,:

```python
        efac = parameter.Constant(1.0)
        equad = parameter.Constant(-np.inf)
        if inc_ecorr:
            ecorr = parameter.Constant(-np.inf)
```

### DIFF
--- a/enterprise/signals/white_signals.py
+++ b/enterprise/signals/white_signals.py
@@ -54,12 +54,18 @@ def WhiteNoise(varianceFunction, selection=Selection(selections.no_selection), n
 
 @function
 def efac_ndiag(toaerrs, efac=1.0):
-    return efac**2 * toaerrs**2
+    if efac is None:
+        return toaerrs**2
+    else:
+        return efac**2 * toaerrs**2
 
 
 @function
 def combined_ndiag(toaerrs, efac=1.0, log10_t2equad=-8):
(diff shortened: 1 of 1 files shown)


vallis:
These functions expect numerical values, and they have even defaults for them. I am not sure it's appropriate to give them flow branches to handle `None`. (Doing so, among other things, may prevent code compilation in some frameworks.) I would prefer a fix in e_e.

mattpitkin:
@vallis No problem. I'll close this and submit a PR for enterprise_extensions.

PR #382, opened 2024-04-17
Change ssb_to_psb_xyz_ICRS to not pass PINT a quantity

tcromartie:
See https://github.com/nanograv/PINT/issues/1733

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -23,9 +23,7 @@ logger = logging.getLogger(__name__)
 try:
     import libstempo as t2
 except ImportError:
-    logger.warning(
-        "libstempo not installed. Will use PINT instead."
-    )  # pragma: no cover
+    logger.warning("libstempo not installed. Will use PINT instead.")  # pragma: no cover
     t2 = None
 
 try:
@@ -34,9 +32,7 @@ try:
     from pint.residuals import Residuals as resids
     from pint.toa import TOAs
 except ImportError:
-    logger.warning(
-        "PINT not installed. Will use libstempo instead."
-    )  # pragma: no cover
+    logger.warning("PINT not installed. Will use libstempo instead.")  # pragma: no cover
     pint = None
 
 if pint is None and t2 is None:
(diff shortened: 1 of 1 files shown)


vhaasteren:
Thank you for fixing this @tcromartie. I just looked at this, and I'm not convinced Enterprise should be changed. As I said in the [PINT issue](https://github.com/nanograv/PINT/issues/1733), PINT is inconsistent between Equatorial and Ecliptic coordinates. PINT should be accepting units, and Enterprise is providing them. Why would the Ecliptic coordinate model accept units, and the Equatorial coordinate model not accept them? Wherever possible, PINT should be units-aware.

enterprise/pulsar.py
```diff
@@ -367,7 +367,7 @@ def __init__(self, toas, model, sort=True, drop_pintpsr=True, planets=True):
             "AstrometryEquatorial" if "AstrometryEquatorial" in model.components else "AstrometryEcliptic"
         )
 
-        self._pos_t = model.components[which_astrometry].ssb_to_psb_xyz_ICRS(model.get_barycentric_toas(toas)).value
+        self._pos_t = model.components[which_astrometry].ssb_to_psb_xyz_ICRS(model.get_barycentric_toas(toas).value)
```
vhaasteren, line 370: Add at the top:

`from astropy.time import Time`

Then change this line to:

`self._pos_t = model.components[which_astrometry].ssb_to_psb_xyz_ICRS(Time(model.get_barycentric_toas(toas),format='mjd')).value`

vhaasteren (changes requested):
Thanks!

Change according to review and I'll merge

enterprise/pulsar.py
tcromartie: Seems good, will do this afternoon! Sorry for the delay

tcromartie: @vhaasteren I thought my second commit was failing because I didn't run black, but it looks like that induced way more changes. I didn't see anything official about this in the contributing guidelines. What are best practices here?

scottransom: Definitely need black!  But can't be the latest version, I think.... Check the dev requirements?  I'm on my phone and will look after I respond here...

scottransom: Yeah.  v23.

tcromartie: Thanks so much Scott! I actually ran test/lint/docs this time (with correct black ver) so hopefully it'll be ok now :)

vhaasteren: Yeah, some of the old committed code was from before black. So if you run black on the whole file yourself, you re-format more than just what you changed. Not very neat. But it' not wrong to put it in, let's just commit it.

PR #387, opened 2024-06-18
Get TM processes from ConditionalGP

blarsen10:
This PR allows you to input a list of timing model parameters to `ConditionalGP` and return the realizations of those individual timing model processes, alongside all other processes which use a basis. This is useful because we can now isolate realizations from individual timing processes (such as spindown or dispersion) from the rest of the timing model. The only catch is users will need to also input the original `Pulsar` object to identify timing parameter name (but you need this to make the `PTA` object anyways).


blarsen10:
Shout out `la_forge.gp` where this functionality is taken from

blarsen10:
Could someone check if this is okay to merge? Maybe @AaronDJohnson or @vallis?

vhaasteren (approved):
All good as far as I can see

vhaasteren:
Hi @blarsen10, sorry that this took time to review. Sometimes dev work can be on the backburner.

It looks fine, and I'm happy to merge. Do I see right that you have also tested this when sampling the coefficients explicitly?

blarsen10:
Thanks for looking at this @vhaasteren ! Glad you asked about the coefficients, I didn't test that part since I was interested in looking at the timing model realizations in the time domain. I realize now that I didn't update it correctly, it's currently adding a new key for every new TM parameter, but duplicating the set of all coefficients from the linearized timing model, instead of only outputting only the coefficient for that parameter. I'll update this so it works as expected and let you know when its ready to merge.

I'll add that I don't think the functionality to sample individual TM keys is strictly necessary for coefficients, since that information can already by isolated out when sampling the coefficients from the whole model (unlike the time-domain processes), but adding coefficients for the individual TM keys to the dictionary at least has the benefit that it will do the key-matching for you already.

blarsen10:
This should be good to go now

PR #366, opened 2023-11-15
adding a hypercube transform function

astrolamb:
With @vhaasteren's ppf additions to `parameter.py`, I've written a short function to transform from a unit hypercube to the prior space for nested sampling

PR #389, opened 2024-07-12
PPF hypercube

astrolamb:
I forgot to make a PR of my fixes to PR #366. 

I'm adding a method to the `PTA` object to compute the PPF of a parameter sampled from a unit hypercube. This allows the user to easily use enterprise `PTA` objects in nested samplers.

There is also a unit test for the method.


vhaasteren (approved):
This all looks fine to me

vhaasteren:
Hi @astrolamb , there are two files that have changes related to twine secrets and bumping the version. That should not be included in a PR like this one. Can you remove those?

.github/workflows/ci_test.yml
```diff
@@ -132,5 +132,6 @@ jobs:
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        TWINE_REPOSITORY_URL: https://upload.pypi.org/legacy/
```
vhaasteren, line 135: This one should not be in this PR

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.2.2
+current_version = 3.4.1
```
vhaasteren, line 2: This should not be in this PR

vhaasteren (changes requested):
Strip out unrelated changes

astrolamb:
Apologies @vhaasteren, I'm not sure why those edit were in there. I've reverted those changes

vhaasteren:
Great. Now you just need to do some linting:

```
would reformat tests/test_vector_parameter.py
would reformat enterprise/signals/signal_base.py
```

Perhaps best is not to let black re-format the whole file, but only the lines that you changed in this PR. You can give black a line range, or you can just manually merge in the reformatting of black. There are more complicated things to do, but above is what I typically do

astrolamb:
@vhaasteren linting done!

PR #400, opened 2024-09-24
feat: partial caching of chromatic Fourier design matrix

jeremy-baier:
This PR adds the ability to cache portions of the construction of the chromatic Fourier design matrix. 
Initially testing on NG15yr data shows a 10x speed up to the recalculation of the chromatic Fmat construction.
This constitutes anywhere between a 3% to 15% speed up of the overall likelihood calculation for single pulsar noise when sampling in the chromatic index, alpha.

As the function docstrings say, the actual caching occurs via the `@function` decorator, which now wraps the later function, caching the initially passed arguments which include the achromatic portion of the Fourier design matrix and the division of TOA observational frequencies by a reference frequency.

This does not play well with the custom noise blocks in `enterprise_extensions` which generically set up chromatic noise as such it can only be set up on a per-pulsar basis.

### DIFF
--- a/enterprise/signals/gp_bases.py
+++ b/enterprise/signals/gp_bases.py
@@ -334,7 +334,7 @@ def construct_chromatic_cached_parts(
 
 @function
 def createfourierdesignmatrix_chromatic_with_additional_caching(fmat_red=None, Ffreqs=None,
-                                                                fref_over_radio_freqs=None, alpha=4.0):
+                                                                fref_over_radio_freqs=None, idx=4.0):
     """
     Construct Scattering-variation fourier design matrix with a cached achromatic component of the
     Fourier design matrix (fmat_red). (Note this is independent of the actual achroamtic basis.)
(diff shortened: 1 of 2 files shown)


AaronDJohnson:
@jeremy-baier linting failed. Can you run the linter? I think `enterprise` may be on a slightly older version of flake8/black. So just check to see what version it is using..

jeremy-baier:
@AaronDJohnson looks like everything is passing now except one part of the code cov. Does that just mean I have to make a test for the 2 added functions?

AaronDJohnson:
@jeremy-baier codecov measures the number of lines covered by tests. So more tests will increase codecov % as long as those lines that you're covering are not already tested.

jeremy-baier:
@AaronDJohnson I added a test which covers both of the added functions. This seemed to satisfy the `codecov` bot.
I think this PR is ready for merge now!

jeremy-baier:
marking this as a draft for a second because I need to check one part of the compatibility with `enterprise_extensions`

vhaasteren:
@jeremy-baier , I marked it as a draft by just using the feature. Or did you by now verified this is compatible?

jeremy-baier:
checking back in on this @AaronDJohnson @vhaasteren . let me know what makes sense !

vhaasteren:
Hi @jeremy-baier , if it isn't compatible with with the enterprise_extensions chromatic noise block, and it is not called by anything else inside Enterprise... is it then perhaps more appropriate to commit this to enterprise_extensions? Then the compatibility/non-compatibility is in one-place. Makes maintenance easier, and it seems that doing non-standard things is more of an extensions kind-of-thing.

jeremy-baier:
I can move this to `enterprise_extensions`. I already have a PR open there for some chromatic noise stuff.

vhaasteren:
Hi @jeremy-baier, from the looks of it, I think that would be best. Thanks for the good work!

vhaasteren:
Closing, because moving to enterprise_extensions

PR #404, opened 2024-12-09
Switched test_orf to nside=16 to decrease memory footprint

vhaasteren:
This decreases the memory footprint of `test_orf` so that github actions may run on it.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -214,7 +214,7 @@ class TestUtils(unittest.TestCase):
         anis_basis = anis.anis_basis(psr_positions, lmax=1, nside=8)
         anis_orf = round(utils.anis_orf(p1, p2, [0.0, 1.0, 0.0], anis_basis=anis_basis, psrs_pos=[p1, p2], lmax=1), 3)
         #anis_orf_exp = -0.150   <-- for nside=16, for nside=8 -0.152
-        anis_orf_exp = -0.150
+        anis_orf_exp = -0.152
         #
 
         msg = "ORF cross term incorrect for {}"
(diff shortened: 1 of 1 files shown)


vhaasteren:
Alright, I guess that doesn't work :/

vhaasteren:
Alright, I'm a little lost now. It succeeds for macos-13 python3.8, but it fails for the other macos-python versions

AaronDJohnson:
@vhaasteren is the Cython version different for 3.8? It was this Cython update that caused the whole issue in the first place as far as I can tell. Or perhaps it coincided with Numpy 2?

vhaasteren:
> @vhaasteren is the Cython version different for 3.8? It was either this Cython or Numpy 2 update that caused the whole issue in the first place, as far as I can tell.

Ah, now there's something! Good catch. The macos py3.8 version lists:

```
Collecting numpy>=1.16.3 (from enterprise-pulsar==0.1.dev1+g9af376a)
  Using cached numpy-1.24.4-cp38-cp38-macosx_10_9_x86_64.whl.metadata (5.6 kB)
```
  
  And the macos py3.9 version says:
  
```
  Collecting numpy>=1.16.3 (from enterprise-pulsar==0.1.dev1+g9af376a)
  Using cached numpy-2.0.2-cp39-cp39-macosx_10_9_x86_64.whl.metadata (60 kB)
```

So the latter one uses numpy-2.02. You are right, that could be it

vhaasteren:
It's odd. On my M1 arm environment, test_utils passes all tests with numpy 2.2.1 and healpy 1.18

vhaasteren:
Not sure how to address this.

PR #409, opened 2025-01-30
Update pulsar.filter_data

blarsen10:
- Fixes a missing underscore, which previously caused planetssb to not update properly
- Allows option to input custom mask if not doing a simple time-slice
- Improves docstring

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -135,12 +135,19 @@ class BasePulsar(object):
             self._isort = slice(None, None, None)
             self._iisort = slice(None, None, None)
 
-    def filter_data(self, start_time=None, end_time=None):
-        """Filter data to create a time-slice of overall dataset."""
-        if start_time is None and end_time is None:
-            mask = np.ones(self._toas.shape, dtype=bool)
-        else:
-            mask = np.logical_and(self._toas >= start_time * 86400, self._toas <= end_time * 86400)
+    def filter_data(self, mask=None, start_time=None, end_time=None):
+        """
+        Filter data to create a time-slice of overall dataset.
+        params:
+            mask : input custom filter (array of bools). None to use start_time and end_time. default None.
+            start_time : MJD of first TOA to include in timeslice. default None. ignored if mask is not None
+            end_time : MJD of last TOA to include in timeslice. default None. ignored if mask is not None
+        """
+        if mask is None:
+            if start_time is None and end_time is None:
+                mask = np.ones(self._toas.shape, dtype=bool)
+            else:
+                mask = np.logical_and(self._toas >= start_time * 86400, self._toas <= end_time * 86400)
 
         self._toas = self._toas[mask]
         self._toaerrs = self._toaerrs[mask]
@@ -158,7 +165,7 @@ class BasePulsar(object):
                 self._flags[key] = self._flags[key][mask]
 
         if self._planetssb is not None:
-            self._planetssb = self.planetssb[mask, :, :]
+            self._planetssb = self._planetssb[mask, :, :]
 
         self.sort_data()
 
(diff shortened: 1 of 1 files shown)


enterprise/pulsar.py
```diff
@@ -135,12 +135,21 @@ def sort_data(self):
...
+        Filters the dataset to create a time-slice based on a custom mask or time range.
+
+        Parameters:
+            mask (array-like, optional): Boolean array specifying which data to keep.
+                                         If None, `start_time` and `end_time` are used. Default is None.
+            start_time (float, optional): Start time (MJD) for filtering. Ignored if `mask` is provided. Default None.
+            end_time (float, optional): End time (MJD) for filtering. Ignored if `mask` is provided. Default None.
+        """
+        if mask is None:
+            if start_time is None and end_time is None:
```
vhaasteren, line 149: Maybe make this if-statement flat?

So

```python
if mask is not None:
    # do stuff
elif start_time is not None or end_time is not None:
    start_time is start_time if start_time is not None else np.min(self._toas)
    end_time is # etc
else:
    # mask = np.ones(self._toas.shape, dtype=bool)
```

vhaasteren:
Hi @blarsen10 ,

Can you merge dev or master and push your branch? If you then also address above comment, I will merge it.

Thanks!

enterprise/pulsar.py
vhaasteren: @blarsen10 

To avoid having to re-write tests, you could get rid of if-statements by starting with 

```python
start_time is start_time if start_time is not None else np.min(self._toas)
mask = np.ones(self._toas.shape, dtype=bool)
mask[self._toas < start_time] = False
# etc
```

vhaasteren:
Looks like this function was not covered in tests in the first place, so number of lines tested stays the same. I'm merging

enterprise/pulsar.py
```diff
@@ -135,12 +135,20 @@ def sort_data(self):
...
+        Parameters:
+            mask (array-like, optional): Boolean array specifying which data to keep.
+                                         If None, `start_time` and `end_time` are used. Default is None.
+            start_time (float, optional): Start time (MJD) for filtering. Ignored if `mask` is provided. Default None.
+            end_time (float, optional): End time (MJD) for filtering. Ignored if `mask` is provided. Default None.
+        """
+
+        start_time = start_time * 86400 if start_time is not None else np.min(self._toas)
+        end_time = end_time * 86400 if end_time is not None else np.max(self._toas)
+        mask = mask if mask is not None else np.logical_and(self._toas >= start_time, self._toas <= end_time)
```
vhaasteren, line 151: Do we want to combine, with a logical and, the mask that was given, and the mask from start_time and end_time?

vhaasteren (commented):
I'm happy to merge. But, perhaps we want to consider whether we want to be able to combine mask and start_time and end_time?

vhaasteren:
Hi @blarsen10 ,

I would propose to modify it to something like the below. Then the and/or in the docstring makes sense:

```python
def filter_data(self, mask=None, start_time=None, end_time=None):
    """
    Filters the dataset to create a time-slice based on a custom mask and/or a time range.
    
    Parameters:
        mask (array-like, optional): Boolean array specifying which data to keep.
                                     If provided, this mask will be combined (logical AND)
                                     with the time range filter. Default is None.
        start_time (float, optional): Start time (MJD) for filtering. If None, the minimum time in the dataset is used.
        end_time (float, optional): End time (MJD) for filtering. If None, the maximum time in the dataset is used.
    """
    
    start_time = start_time * 86400 if start_time is not None else np.min(self._toas)
    end_time = end_time * 86400 if end_time is not None else np.max(self._toas)
    mask_times = np.logical_and(self._toas >= start_time, self._toas <= end_time)
    mask = np.logical_and(mask, mask_times) if mask is not None else mask_times
```

enterprise/pulsar.py
blarsen10: Yes that's a good idea, I updated this using your suggestion below!

PR #411, opened 2025-02-26
Linear interpolated FFT-based time-correlated signals

vhaasteren:
This PR introduces the linear interpolated FFT-based rank reduced time-correlated signals. These replace the diagonal Phi matrix approach that Enterprise has traditionally used based on the Lentati et al. (2013) method.

Using it is pretty simple. Instead of:

```python
rn_pl = powerlaw(log10_A=rn_log10_A, gamma=rn_gamma)
rn_phi = gp_signals.FourierBasisGP(spectrum=rn_pl, components=n_components, Tspan=Tspan)
```

one would do:

```python
rn_pl = powerlaw(log10_A=rn_log10_A, gamma=rn_gamma)
rn_fft = gp_signals.FFTBasisGP(spectrum=rn_pl, components=n_components, oversample=3, cutoff=3, Tspan=Tspan, start_time=start_time)
```

so the same spectral function can be used. Free spectrum is NOT supported yet. Instead of `components`, we can also pass `knots=`, where it is understood that `knots=2*n_components+1`. This is because `components` actually means frequencies. In the time-domain, the number of `knots` is the number of `modes+1`, because we cannot just omit the DC term.

The `oversample` parameter determines how densely the PSD is sampled in frequencies. With `oversample=1` we would use frequencies at spacing of `df=1/T`. With `oversample=3` (the default), the frequency spacing is `df=1/(3T)`. Note that this is a way to numerically approximate the Wiener-Khinchin integral. With oversample sufficiently large, the FFT is an excellent approximation of the analytical integral. For powerlaw signals, `oversample=3` seems a very reasonable choice.

The `cutoff` parameter is used to specify below which frequency `fcut = 1 / (cutoff*Tspan)` we set the PSD equal to zero. Note that this parameterization (which is also in Discovery) is a bit ambiguous, as `fcut` may not correspond to an actual bin of the FFT: especially if `oversample` is not a high number this can cause a mismatch. In case of a mismatch, fcut is rounded up to the next oversampled-FFT frequency bin. Instead of `cutoff`, the parameter `cutbins` can also be used (this overrides `cutoff`). With `cutbins` the low-frequency cutoff is set at: `fcut = cutbins / (oversample * Tspan)`, and its interpretation is less ambiguous: it is the number of bins of the over-sampled PSD of the FFT that is being zeroed out.

For common signals, instead of:

```python
gw_pl = powerlaw(log10_A=gw_log10_A, gamma=gw_gamma)
orf = utils.hd_orf()
crn_phi = gp_signals.FourierBasisCommonGP(gw_pl, orf, components=20, name='gw', Tspan=Tspan)
```

one would do:

```python
gw_pl = powerlaw(log10_A=gw_log10_A, gamma=gw_gamma)
orf = utils.hd_orf()
crn_fft = gp_signals.FFTBasisCommonGP(gw_pl, orf, components=20, name='gw', Tspan=Tspan, start_time=start_time)
```

DM-variations and Chromatic noise can be similarly set up:
```python
nknots = 81
dm_basis = utils.create_fft_time_basis_dm(nknots=nknots)
dm_pl = powerlaw(log10_A=dm_log10_A, gamma=dm_gamma)
dm_fft = gp_signals.FFTBasisGP(dm_pl, basis=dm_basis, nknots=nknots, name='dmgp')

chrom_basis = utils.create_fft_time_basis_chromatic(nknots=nknots, idx=chrom_idx)
chrom_pl = powerlaw(log10_A=chrom_log10_A, gamma=chrom_gamma)
chrom_fft = gp_signals.FFTBasisGP(chrom_pl, basis=chrom_basis, nknots=nknots, name='chromgp')
```

Subtleties
----------

Enterprise allows one to combine basis functions when they are the same. This
is especially useful when analyzing common signals which have the same basis as
a single-pulsar signal, such as one would have with red noise and a correlated
GWB. This can be done with the `combine=True` option in `FFTBasisGP` and
`FFTBasisCommonGP`. Default is `combine=True`. The subtlety is that modern PTA
datasets typically have large gaps, which causes some of the time-domain basis
functions to basically be all zeros. Therefore, some basis functions that you
would not expect to be identical will be combined.

The above is not a bug. Combining such bases and the corresponding Phi matrix
does not matter, because the basis is zero, and that part of the signal has no
bearing on the data or the model. However, when doing signal reconstruction,
such as with `la_forge` or `utils.ConditionalGP`, make sure to set `combine=False`.

Extra comment
---------------

An additional re-design of certain patterns is needed to accelerate TNT construction using sparse matrices for faster single pulsar noise analysis. Because of the severity of changes, that will be a different PR.


enterprise/signals/gp_signals.py
```diff
@@ -216,6 +228,87 @@ class FourierBasisGP(BaseClass):
...
+):
+    """Function to return a BasisGP class with a coarse time basis."""
+
+    if nknots is None:
+        nknots = 2 * components + 1
+
+    elif nknots is not None and nknots % 2 == 0:
+        raise ValueError("Knots needs to be an odd number")
+
+    if cutoff is not None:
```
meyers-academic, line 256: You say in the comments and the docs that `cutbins` overrides `cutoff` but here it looks like the reverse to me (i.e. `cutbins` is used unless `cutoff` is set, in which case it uses `cutoff`.

enterprise/signals/gp_signals.py
```diff
@@ -216,6 +228,87 @@ class FourierBasisGP(BaseClass):
...
+        def _construct_prior(self, params):
+            for key, slc in self._slices.items():
+                t_knots = self._labels[key]
+
+                freqs = utils.knots_to_freqs(t_knots, oversample=oversample, fmax_factor=fmax_factor)
+
+                # Hack, because Enterprise adds in f=0 and then calculates df,
+                # meaning we cannot simply start freqs from 0. Thus, we use
+                # a modified frequency spacing, such that:
+                # [0, f1, 0, f1, f2, f3] => [df, -df, df, df, df]
```
meyers-academic, line 285: I'm not sure I understand this. So *later* f=0 gets added? So what you construct here appears to be `[f1, 0, f1, f2, f3]` and somewhere else f=0 gets added? But I'm not sure I understand why this `-df` that you include here applied to f1 helps. 

Additionally, down below you have `psd = [[-psd_prior[1], psd_prior[2:]]` which now to me looks like it has fewer elements than freqs_prior because `psd_prior` should have the same number of elements as `freqs_prior` so psd looks like it will have one less element. Is that because `psd` should match `freqs` because it gets passed to `psd2cov`? 

I'm just a bit confused by what this code does. because `psd`, if I've read everything right seems to have [-psd(f=0), psd(f=f1), ...] in it. So is the point that if `psd(f=0)` gets added somewhere else what is here will cancel with that when summed over?

enterprise/signals/gp_signals.py
```diff
@@ -173,10 +187,8 @@ def get_basis(self, params={}):
 
             def get_phi(self, params):
                 self._construct_basis(params)
+                self._construct_prior(params)
 
-                for key, slc in self._slices.items():
-                    phislc = self._prior[key](self._labels[key], params=params)
-                    self._phi = self._phi.set(phislc, slc)
```
meyers-academic, line 179: Ok, so this is deleted because this `_phi` is set in `construct_prior` now?

Also, as an aside "_labels" is a terrible name for the frequencies/times associated with one of the axes of the basis. But I say that with absolutely no suggested replacement...but it took me a while to find what it actually meant :).

vhaasteren:
Hi @meyers-academic , do you think you can have a look at this PR? I had not yet made the knots->nodes change until now, but otherwise this PR has been ready for a while

enterprise/signals/gp_signals.py
meyers-academic: just reiterating this here -- this confusion still exists in the docs/code.

meyers-academic:
Ok -- there are just the two questions still above about cutoff vs. cutbins, and the internal structure of the frequency arrays. Could you say something about that maybe? At least for posterity's sake?

the tests seem sensible for checking that the inner workings make sense.

meyers-academic (commented):
Ok, Hopefully now my comments will show up.

enterprise/signals/gp_signals.py
vhaasteren: Yes, that is correct. We can now cache `prior_params`, just like `white_params`, `basis_params`, and `delay_params` this way.

vhaasteren: And noted that `_labels` is a HORRIBLE name for frequencies. It's such a shame, but that's simply the only way the frequency information is stored in Enterprise.

enterprise/signals/gp_signals.py
vhaasteren: Yes @meyers-academic , as I comment in the code, this is ugly. This is unfortunately a design flaw in Enterprise. All prior functions are implemented as:

```python
@function
def turnover(f, log10_A=-15, gamma=4.33, lf0=-8.5, kappa=10 / 3, beta=0.5):
    df = np.diff(np.concatenate((np.array([0]), f[::2])))
    hcf = 10**log10_A * (f / const.fyr) ** ((3 - gamma) / 2) / (1 + (10**lf0 / f) ** kappa) ** beta
    return hcf**2 / 12 / np.pi**2 / f**3 * np.repeat(df, 2)
```

See that `df` that has no business being in a PSD function? When setting `components=1`, or going to `f=0`, like we need to, the function does not work, because df becomes zero. See how `df` is calculated by first adding `f=0` in the front of the array for `f`? Enterprise should have left the PSD (the prior) just calculate only that, just the PSD. The addition of `df` in the prior function violates a couple of programming principles. However, solving it is a major refactor with many user-facing breaking changes. That is far outside the scope of this PR

This was the only way I could think of to get an array for the PSD at values for f starting at zero, with components=1

vhaasteren: Note that I wrote a separate PR that makes that hard-coded `2` in the `::2` a variable called `components`. That's how it is in other prior functions (PSD functions), so it is inconsistent that it is hardcoded in only a subset of them. It's in: #418

PR #412, opened 2025-03-05
Dev to Master

vhaasteren:
There have been quite a few updates again. Most significantly, no more dependency on libstempo and PINT, which will help with dependency-hell for other packages.

Changes:

- Adding hypercube transformations and PPF functions
- Support for feather files
- Removal of libstempo and PINT as mandatory dependencies
- Bugfix: allow 2D left / right arrays in MarginalizingNmat
- Updated examples


enterprise/signals/utils.py
```diff
@@ -93,12 +95,27 @@ def _sample_conditional(self, params, n=1, gp=False, variance=True):
...
+                        if "timing_model" in sig.name and len(self.tm_params) > 0:
+                            if self.psr is None:
+                                raise ValueError("Need to input psr to get timing model param names")
+                            else:
+                                for tm_par in self.tm_params:
+
+                                    tm_idx = list(self.psr.fitpars).index(tm_par)
+                                    save_name = sig.name.split("_")[0] + "_" + tm_par
+
+                                    if gp:
```
AaronDJohnson, line 114: We're 8 indentations deep at this point. Can it be cleaned up at all?

AaronDJohnson (changes requested):
There is only one section that could use a little cleaning for maintainability. If we want to do it later, I will submit an issue instead.

AaronDJohnson:
The rule sets have been changed to accept mac-os latest, so if we rerun the CI, we should be able to merge.

enterprise/signals/utils.py
vhaasteren: I cleaned it up a bit with the philosophy "fail fast"

vhaasteren: I cleaned it up a bit with the philosophy "fail fast"

PR #416, opened 2025-03-27
Feather metadata

vhaasteren:
Bugfix: update featherpulsar metadata


AaronDJohnson (commented):
Looks fine to me. I disagree with the codecov patch settings. It causes problems for quick bug fixes like this which are only 3 lines long.

PR #418, opened 2025-04-09
Fix prior components

### LINKED ISSUE #417: Components missing in several gp_prior functions
In gp_priors we define priors like so:

```python
@function
def powerlaw(f, log10_A=-16, gamma=5, components=2):
    df = np.diff(np.concatenate((np.array([0]), f[::components])))
    return (
        (10**log10_A) ** 2 / 12.0 / np.pi**2 * const.fyr ** (gamma - 3) * f ** (-gamma) * np.repeat(df, components)
    )
```

The `components` parameter defaults to 2, because we have two modes per frequency. However, some functionality in Enterprise requires the user to set this number to something else. The recent FFT signals are an example, but the functionality has existed well before. In some GP priors those are missing, such as:

```python
@function
def broken_powerlaw(f, log10_A, gamma, delta, log10_fb, kappa=0.1):
    """
    Generic broken powerlaw spectrum.
    :param f: sampling frequencies
    :param A: characteristic strain amplitude [set for gamma at f=1/yr]
    :param gamma: negative slope of PSD for f > f_break [set for comparison
        at f=1/yr (default 13/3)]
    :param delta: slope for frequencies < f_break
    :param log10_fb: log10 transition frequency at which slope switches from
        gamma to delta
    :param kappa: smoothness of transition (Default = 0.1)
    """
    df = np.diff(np.concatenate((np.array([0]), f[::2])))
    hcf = (
        10**log10_A
        * (f / const.fyr) ** ((3 - gamma) / 2)
        * (1 + (f / 10**log10_fb) ** (1 / kappa)) ** (kappa * (gamma - delta) / 2)
    )
    return hcf**2 / 12 / np.pi**2 / f**3 * np.repeat(df, 2)

```

That breaks certain functionalities like the FFT signals.

vhaasteren:
This fixes #417

### DIFF
--- a/enterprise/signals/gp_priors.py
+++ b/enterprise/signals/gp_priors.py
@@ -20,10 +20,10 @@ def powerlaw(f, log10_A=-16, gamma=5, components=2):
 
 
 @function
-def turnover(f, log10_A=-15, gamma=4.33, lf0=-8.5, kappa=10 / 3, beta=0.5):
-    df = np.diff(np.concatenate((np.array([0]), f[::2])))
+def turnover(f, log10_A=-15, gamma=4.33, lf0=-8.5, kappa=10 / 3, beta=0.5, components=2):
+    df = np.diff(np.concatenate((np.array([0]), f[::components])))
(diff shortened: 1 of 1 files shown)


AaronDJohnson (approved):
Looks good to me.

PR #419, opened 2025-04-21
Add sqrtN solve for sherman-morrison and white noise kernels

vhaasteren:
Let `N` be the white noise kernel (`marginalizingNmat`, `ShermanMorrison`, `ndarray_alt`, ...). The `solve(x)` method would calculate `N^{-1}x`. Assume that `LL^T=N` with `L` a Cholesky factor. Then the new `sqrtsolve(x)` would return `L^{-1}x`

This PR implements that


vhaasteren:
Hi @AaronDJohnson, any chance you can give this a quick review? My p-value paper is entering production

AaronDJohnson (approved):
Sorry for the delay! This looks good.

PR #423, opened 2025-08-06
Bugfix: add flat powerlaw to gp_priors.py

jeremy-baier:
enterprise_extensions has a flat powerlaw model in the common_red_noise_block which tries to use a GP prior from enterprise, but it does not exist in the main or dev branches. This PR adds that GP prior.

line in enterprise extensions which calls the flat powerlaw:
https://github.com/nanograv/enterprise_extensions/blob/4e1ce072f179ceb572f593e7af15329c3f645d77/enterprise_extensions/blocks.py#L1460

flat powerlaw taken from:
https://gitlab.in2p3.fr/epta/enterprise/-/blob/master/enterprise/signals/gp_priors.py?ref_type=heads

### DIFF
--- a/enterprise/signals/gp_priors.py
+++ b/enterprise/signals/gp_priors.py
@@ -14,7 +14,9 @@ import enterprise.constants as const
 @function
 def powerlaw(f, log10_A=-16, gamma=5, components=2):
     df = np.diff(np.concatenate((np.array([0]), f[::components])))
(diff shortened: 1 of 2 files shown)


AaronDJohnson (approved):
Looks good to me. In the future, we should consider maintainability improvements by creating a base powerlaw function that takes df as an argument. It would reduce code duplication and keep speed the same.

PR #429, opened 2026-02-09
Fix/pkg resources

astrolamb:
Fixing pkg_resources bug in `enterprise.signals.utils.py`. Found by @levischult.

Additionally, I used what I learnt to remove an import to the entirety of `enterprise` in `enterprise.pulsar`

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -74,7 +74,7 @@ class BasePulsar(object):
     def _get_pdist(self):
         path = resources.files("enterprise")/"datafiles/pulsar_distances.json"
         with open(str(path), 'r') as file:
-            pdict = json.load(path)
+            pdict = json.load(file)
 
         if self.name[0] not in ["J", "B"]:
             if "J" + self.name in pdict:
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -73,7 +73,7 @@ class BasePulsar(object):
 
     def _get_pdist(self):
         path = resources.files("enterprise")/"datafiles/pulsar_distances.json"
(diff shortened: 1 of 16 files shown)


AaronDJohnson:
I've never tried the copilot review, so let's see how it does.

vhaasteren:
Thanks for fixing this!

There is a lot of stuff in this PR that is not doing anything. It seems to be linting and other stuff. I know enterprise is not fully linted according to requirements, and we should probably fix that. But not in this PR

Also, why are the tests failing for python 3.11 and 3.12?

astrolamb:
@vhaasteren all tests fail when I don't lint. After some linting, the all tests except for python>=3.11 pass, which I also don't understand. Anyway, that's why there's mostly empty stuff in there

astrolamb:
I'm realising I branched from `master` and not from `dev`, and PRing into `master`. Could that be a reason why we're seeing so many failures?

astrolamb:
Update: numpy>=1.25 had some deprecations that resulted in failed tests. These are now fixed in this PR and should be merged to master

vhaasteren:
@astrolamb 

You can only lint the sections you modify. Also you have edits to gp_priors.py, signal_base.py that are unrelated and should not be present in this PR

Any stuff a linter would modify but that is not modified in this PR will not trigger the github tests. Then the PR would be much neater. I know this is a bit of a pain, but this repo is not enterprise extensions 😉 

Then, in a later PR, we can just lint the entire codebase.

astrolamb:
I have restarted this fix in two others branches, which I will now PR. These are based on `dev`. This has avoided a lot of the linting bugs that I have faced :)

PR #433, opened 2026-02-13
Fix/pkg resources v2

### LINKED ISSUE #428: `pkg_resources` deprecation
While debugging some issues with a new environment, I noticed that `from enterprise.pulsar import Pulsar` raises a `DeprecationWarning`:
```
[...]/enterprise/signals/utils.py:13: UserWarning: pkg_resources is deprecated as an API.
See https://setuptools.pypa.io/en/latest/pkg_resources.html.
The pkg_resources package is slated for removal as early as 2025-11-30.
Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import Requirement, resource_filename
```
Apparently `pkg_resources` (from `setuptools`), which is used in `signals/utils.py`, has been deprecated in favor of the native Python module [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html).

astrolamb:
A cleaner attempt to fix #428!


astrolamb:
Sorry, I PR'd to `master`. Let me put it into `dev`

PR #434, opened 2026-02-13
Fix/pkg resources v2

astrolamb:
A cleaner fix to #428. Replacing #429


astrolamb:
Failed tests are fixed by #435

AaronDJohnson:
Fixed with #434

PR #435, opened 2026-02-13
Fix/numpy deprecations

astrolamb:
Fixing numpy v1.25 depreciations causing unit test failures in python 3.11 and 3.12 - [#10615](https://github.com/numpy/numpy/pull/10615)

This based on PR #433 -- required for unittest passing in python 3.11 and python 3.12.

**I strongly suggest pushing to `master` soon as this will cause installation errors, unless we fix numpy<1.25**

### DIFF
--- a/enterprise/pulsar.py
+++ b/enterprise/pulsar.py
@@ -7,6 +7,7 @@ import json
 import logging
 import os
 import pickle
+from importlib import resources
 
 from pyarrow import feather
 from pyarrow import Table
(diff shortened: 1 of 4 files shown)


vhaasteren:
Yeah, a merge to master will be due soon for quite a few reasons. Thanks for doing the plumbing @astrolamb !

scottransom:
Yes, we need this and #434 merged for TWG work, as well!  So a merge very soon would be great.  Thanks!

PR #446, opened 2026-04-29
Improved version of `make_ecc_interpolant`

abhisrkckl:
Implements a more conservative expression for the number of harmonics needed to accurately represent an eccentric waveform, based on https://arxiv.org/abs/2210.11454.
image

image


vhaasteren:
@abhisrkckl Do we even need this function in enterprise? I thought we removed it because it is dead code...

abhisrkckl:
I didn't realize this was being removed. I will close this PR.

vhaasteren:
@abhisrkckl I _thought_ I remembered we decided that it was being removed, but I couldn't quickly find the discussion...  A quick grep at least shows no enterprise routine uses it.  What would be the right place for this? Enterprise extensions? Or is it a convenience function that we should support?

abhisrkckl:
This is useful for calculating the Fourier expanded eccentric waveform as in Taylor et al. We don't use that for eccentric searches anymore.

